PR #7, opened 2026-05-22
fix: validate admin token before leaving login page

hugalafutro:
## Changes

Previously, entering a wrong admin token would save it to localStorage, reload into the main UI (showing sidebar + spinners), then bounce back to the login page when API calls returned 401.

Now the login page validates the token against `GET /api/system` before saving it:
- **Invalid token** → red "Invalid admin token" error on login page, no navigation
- **Network error** → "Failed to connect to server" error
- **Loading state** → button shows "Signing in..." and is disabled during validation
- **Error cleared on retry** → previous errors disappear when submitting again

Bumps version to 0.9.5.


hugalafutro:
@greptile review

hugalafutro:
@greptile review

PR #33, opened 2026-05-24
fix: live log rows stuck with incomplete data + stream error badge + stats

hugalafutro:
## Summary

Live log rows in scroll mode showed incomplete data (provider="-", status=0, all metrics="-") until page reload. Two additional bugs were discovered during testing and fixed alongside.

### Bug 1: Log rows stuck with incomplete data (main fix)

**Root cause:** The backend inserts a request log row with minimal data (Phase 1: `INSERT` with `state=pending`, `provider_id=NULL`, all metrics zero). When the request completes, it updates the row (Phase 2: `UPDATE` with full metrics). Both phases publish SSE events. The frontend's `useBidirectionalFetch` hook deduplicated by ID — when `request.completed` triggered a re-fetch, the completed row was filtered out because its ID already existed in the list.

**Fix:** Added `GET /api/logs/{id}` endpoint + `api.logs.get(id)` client method. On `request.completed` SSE, the frontend now fetches the single completed entry by ID and merges it into the existing entries in-place via a new `mergeEntries` function on the `useBidirectionalFetch` hook. Falls back to `fetchNewer()` on error.

### Bug 2: Interrupted streams showed green 200 badge

**Root cause:** When a streaming request errored (unexpected EOF, client disconnect, missing `[DONE]` sentinel), the backend kept `status_code=200` from the upstream HTTP response, only setting `errorMessage` and `state="failed"`. The frontend badge logic only considers `status_code`, so failed streams showed a green 200 badge.

**Fix:** Set `logData.statusCode = 0` when `errMsg != ""` in the streaming error path (`proxy.go`). The frontend already treats `statusCode === 0` as error → red badge. Also updated stats queries in `stats.go` to count `status_code=0` as errors and exclude them from success/latency averages.

### Bug 3 (minor): Stats queries ignored `status_code=0` errors

**Root cause:** Stats SQL filtered errors as `status_code >= 400` and success as `status_code >= 200 AND < 400`. Rows with `status_code=0` (now used for stream failures) fell through both buckets, vanishing from dashboards.

**Fix:** Error counts now use `status_code >= 400 OR status_code = 0`. Success/latency filters use `status_code > 0 AND < 400`.

### UI polish (same commit)

- Key column: `whitespace-nowrap` → `max-w-[7rem] truncate` with `title` tooltip
- Counter range: clamped to `entries.length` (was showing impossible ranges like `1–15 / 11`)
- Removed horizontal scrollbar from `VirtualLogTable`

## Files changed

| File | Change |
|------|--------|
| `internal/api/logs.go` | `GET /api/logs/{id}` endpoint (`GetLog` handler) |
| `internal/proxy/proxy.go` | `status_code=0` on stream error |
| `internal/api/stats.go` | Count `status_code=0` as errors in stats queries |
| `web/src/api/client.ts` | `api.logs.get(id)` method |
| `web/src/hooks/useBidirectionalFetch.ts` | `mergeEntries` function |
| `web/src/pages/Logs.tsx` | SSE handler: fetch-by-ID on `request.completed`; key column truncation |
| `web/src/components/VirtualLogTable.tsx` | Counter clamping; overflow-x removal; key column truncation |
| `.gitignore` | `compose.stress-test.yml` |
| `.version` | `0.9.9` → `0.9.10` |


hugalafutro:
Checking the remaining coverage gaps now

PR #50, opened 2026-05-26
fix: eliminate MSW, act(), and jsdom warnings from test suite

hugalafutro:
## Summary

Eliminates all MSW unhandled request warnings, React `act()` warnings, and jsdom "Not implemented" warnings from the test suite (4118 tests, 0 failures, 0 warnings).

### MSW handler gaps
- Add handlers for `/api/logs/cursor`, `/api/models/cursor`, `/api/logs/app/cursor`, `/api/version/latest` in `handlers.ts`
- Add cursor factories (`mockLogsCursor`, `mockAppLogsCursor`, `mockModelsCursor`) to `helpers.ts`

### act() warnings (root cause: MockEventSource)
- Remove `setTimeout(() => { this.onopen?.() }, 0)` from `MockEventSource` constructor — this was the root cause of act() warnings across most test files
- Set initial `readyState` to `OPEN` (1) instead of `CONNECTING` (0)
- Update `infrastructure.test.tsx` readyState assertion accordingly
- Wrap `vi.advanceTimersByTime` in `act()` in `Spinner.test.tsx` and `shared.test.tsx`

### jsdom "Not implemented" warnings
- Add `HTMLElement.prototype.scrollTo` mock in `setup.ts`
- Suppress jsdom virtualConsole `jsdomError` events starting with "Not implemented:" (covers both `scrollTo` and "navigation" warnings)

### RateLimitSettings handler reset
- Change `server.resetHandlers(...)` to `server.use(...)` to preserve the default `/api/events` handler

### Test query disambiguation
- `AppLogs.test.tsx`: `getByText('Level')` → `getByRole('button', { name: 'Level' })`
- `Logs.test.tsx`: `getByText('Status')` → `getByRole('button', { name: 'Status' })`

### DIFF
--- a/web/src/components/__tests__/Spinner.test.tsx
+++ b/web/src/components/__tests__/Spinner.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen } from "@testing-library/react";
+import { act, render, screen } from "@testing-library/react";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { AllProviders } from "../../test/utils";
 import { Spinner } from "../Spinner";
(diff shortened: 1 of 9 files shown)


hugalafutro:
Addressing both Greptile review comments in ff7256e:

**P1: MockEventSource never fires `onopen`** — Fixed. The `MockEventSource` now fires `onopen` via `queueMicrotask` instead of never firing it. `queueMicrotask` runs after the current synchronous block (so callers can set `es.onopen = handler` before the callback fires) but within React 18's `act()` scope, so no act() warnings. Added a dedicated test in `infrastructure.test.tsx` that verifies `onopen` is called when a handler is assigned.

**P2: `_virtualConsole` is private jsdom API** — Investigated both public API alternatives:
1. **`testEnvironmentOptions.virtualConsole`**: Fails because `VirtualConsole` (an `EventEmitter`) cannot be serialized across Vitest's forked worker boundary — `"() => { ... } could not be cloned"`.
2. **`console.error` wrapping**: Fails because jsdom's `VirtualConsole.forwardTo()` forwards `jsdomError` events to the Node.js `console.error`, not the jsdom `window.console.error`. In Vitest's jsdom env, `globalThis === window`, so we can't access the Node.js console from the setup file.

The `_virtualConsole.emit` patch is the only reliable interception point. Added documentation explaining why the alternatives fail, and the `if (win._virtualConsole)` guard ensures graceful degradation if jsdom removes the property (warnings reappear instead of crashing).

PR #103, opened 2026-06-02
Add pnpm-workspace.yaml to fix workspace boundary

hugalafutro:
Adds `pnpm-workspace.yaml` at the project root with `packages: ["web"]`. Without this, pnpm walks up the directory tree and finds `~/pnpm-workspace.yaml`, treating the entire home directory as the monorepo root and installing packages to `~/node_modules/` instead of `web/node_modules/`. Also adds `node_modules/` to `.gitignore` for the root-level virtual store.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,4 @@ Thumbs.db
 .windsurfrules
 .omo
 web/.pnpm-store/
+node_modules/
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 
 *~
 AGENTS.md
+pnpm-workspace.yaml
 *.bak
 /bin/
 # Binaries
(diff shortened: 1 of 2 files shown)


hugalafutro:
Closing. Committing `pnpm-workspace.yaml` breaks CI because pnpm then expects `pnpm-lock.yaml` at the project root instead of `web/pnpm-lock.yaml`, causing `pnpm install --frozen-lockfile` to fail in Docker builds.

This file is only needed on machines where `~/pnpm-workspace.yaml` exists (which makes pnpm walk up and treat the home directory as a workspace root). Most developers won't have that file. For those who do, creating `pnpm-workspace.yaml` locally with `packages: ["web"]` fixes the boundary without affecting CI.

PR #125, opened 2026-06-05
Security hardening: rate limiting, XFF fix, non-root container, error sanitization

hugalafutro:
## Summary

Security hardening addressing findings from a structured threat model analysis. Each commit is a single logical fix.

## Changes

| Commit | Area | Description |
|--------|------|-------------|
| `587d14f1` | HTTP server | Add `ReadHeaderTimeout: 10s` to `http.Server` — prevents slowloris attacks. The `//nolint:gosec` comment claimed this was set but it was missing. |
| `61f36d72` | Rate limiting | Mount IP rate limiter middleware on `/api` and `/api/chat` routes — previously only applied to `/v1/` proxy routes, leaving admin auth endpoints unprotected from brute-force. |
| `3d0426b8` | IP extraction | Walk XFF chain right-to-left, returning the rightmost non-trusted entry instead of blindly trusting the leftmost. Prevents IP spoofing by clients behind a trusted reverse proxy. |
| `e96fb263` | IP extraction | Fix IPv6 edge case — use `net.ParseIP` directly instead of `ip+":0"` hack, which corrupted IPv6 addresses with trailing `::`. |
| `749ec4d2` | Proxy | Sanitize upstream error response in failover-exhausted path — previously forwarded raw upstream JSON (potentially leaking provider request IDs, internal paths). |
| `91108ef8` | SSE | Stop infinite reconnect loop on 401 — clear stored admin token and reload page instead of exponential-backoff looping with stale credentials. |
| `558272e6` | Backup | Strip password from `pg_dump`/`pg_restore` command-line arguments — password now only passed via `PGPASSWORD` env var, not visible in `ps`. |
| `e1dab362` | Dockerfile | Run container as non-root user (uid 1000) with privilege-dropping entrypoint. Surgical `chown` on `/data` and `/data/backups` only — does not touch `pgdata`. Transparent upgrade path for existing deployments. |

## Tests

- **Backend**: 18 packages, all pass. New tests for XFF rightmost-walk (including IPv6), spoofing prevention, multi-hop chains.
- **Frontend**: 94.2% statement coverage, 95.3% line coverage, all pass.

## Deployment note

The Dockerfile change is backwards-compatible. The entrypoint runs as root to fix `/data` ownership, then drops to uid 1000 via `su-exec`. Existing deployments where `.data/` was created as `root:root` by Docker get fixed transparently on first boot.


hugalafutro (commented):
All three findings addressed in `0c122fc7` and test coverage added in `1c2fb0b0`:\n\n1. **P1 rate limiter ordering**: Fixed — `ipLimiter.Middleware` now runs before `AuthMiddleware` on `/api/chat` (line 315).\n\n2. **P1 upstream error passthrough**: Fixed — generic error only returned when `!hasMoreCandidates` (truly exhausted). Non-failover-eligible errors forward the upstream JSON body so clients can react to semantic errors like `context_length_exceeded`.\n\n3. **P2 XFF leftmost fallback**: Fixed — `net.ParseIP(candidate)` check added before returning the leftmost entry. Non-IP strings like `"unknown"` no longer become rate-limiter bucket keys.

hugalafutro:
@greptile-apps — all review findings addressed:

**P1 (rate limiter ordering)**: Fixed in `0c122fc7` — `ipLimiter.Middleware` runs before `AuthMiddleware` on `/api/chat`.

**P1 (generic error swallows all upstream failures)**: Fixed in `0c122fc7` — generic error only when `!hasMoreCandidates`. Non-failover-eligible errors forward upstream JSON.

**P1 (non-JSON upstream passthrough)**: Fixed in `9fb24d0f` — non-JSON bodies wrapped in OpenAI-compatible envelope via `writeOpenAIError`. Only valid JSON forwarded as-is.

**P2 (XFF leftmost fallback)**: Fixed in `0c122fc7` — `net.ParseIP` check before returning leftmost entry.

PR #138, opened 2026-06-06
fix: verified crowdin config

hugalafutro:
Replaces the merged crowdin.yml with the format verified by the CrowdIn CLI.

- Quoted keys throughout
- `base_path: ".." ` to resolve from repo root
- No leading slashes on source/translation paths
- Removed `excluded_target_languages` (not needed)
- **Tested locally**: upload sources ✔, upload translations ✔, download translations ✔


hugalafutro:
mark my words - it won't work

PR #144, opened 2026-06-06
fix: CrowdIn CLI flags and upload_translations

hugalafutro:
Two fixes for the CrowdIn workflow:

1. Removed `skip_untranslated_strings` — CrowdIn CLI does not allow it together with `skip_untranslated_files`. Using files-only skip is sufficient since `returnEmptyString: false` in i18next handles individual empty keys.

2. Disabled `upload_translations` — it tries to push all locale files (including en.json) back to CrowdIn, causing 'Project doesn't have en language(s)' errors. Source upload already handles en.json.

### DIFF
--- a/.github/workflows/crowdin.yml
+++ b/.github/workflows/crowdin.yml
@@ -29,14 +29,11 @@ jobs:
         with:
           # Upload source strings when en.json changes on master
           upload_sources: true
-          # Upload existing translations so Crowdin has the latest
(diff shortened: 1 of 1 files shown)


hugalafutro:
I thought we'd have flying cars and I'd work as an asteroid miner in the future instead I got back pain and bots rolling around my git history like their puppies that found fresh shit in the grass.

PR #148, opened 2026-06-06
fix(proxy): core proxy/streaming review findings

hugalafutro:
Addresses all findings from the core proxy/streaming code review (`plans/core-proxy-streaming-review.md`).

## Fixes

| # | Finding | Commit | Description |
|---|---------|--------|-------------|
| 1 | Breaker tripped by per-model 404 | `a00434f7` → `58ac3d07` → `ec774eaa` | 404/499 → true no-op (neither failure nor success); 401/403 → failure; 429/5xx → failure. Extracted `breakerRecordAction()` with 16-case table test. |
| 2 | Metering dropped on disconnect | `c9a598db` | Removed `!clientDisconnected` gate — both streaming and non-streaming paths now meter consistently. Integration test (real DB + upstream + mid-stream cancel) confirms. |
| 3 | Blind `stream_options` injection | `a6dc500e` | Moved injection into per-candidate rewrite block, gated by `providerSupportsStreamOptions()` allowlist. Anthropic/Google/Cohere never receive it. |
| 4 | No overall failover deadline | `4cde6cbc` | Adds `overallDeadline = startTime + 2×failoverTimeout`, checked at loop top with `attempt > 0`. First candidate gets full budget. |
| 5 | Interim log blocks TTFB | `80a823cd` → `9e59f805` | All `updateRequestLog` calls before response write now skip `WaitForInsert`. Only the streaming final update (off hot path) still waits. |
| 6 | SafeDialer resolver-error bypass | `7b0e7245` | DNS resolution failure returns error directly instead of falling through to unchecked dial. Invariant restored: every dial is IP-checked. |

## Tests

- `breaker_action_test.go` — 16-case table test of `breakerRecordAction()` (the authoritative status→action mapping)
- `circuitbreaker_404_test.go` — 3 tests for no-op preservation of failure history and half-open state
- `helpers_test.go` — `providerSupportsStreamOptions` table test (5 false, 11 true, unknown default)
- `token_usage_test.go` — integration test: real VK + upstream + mid-stream cancel → `TokensUsed==15` in DB
- `safe_dialer_test.go` — DNS error rejection + no bypass of IP blocklist

Full suite passes (18 packages).


hugalafutro:
@hermes-ds4pro hello

PR #161, opened 2026-06-07
refactor(proxy): decompose handleStreamingResponse into a streaming pipeline

hugalafutro:
## What

Behavior-preserving decomposition of the SSE streaming god-function. `handleStreamingResponse` goes from **~965 → 211 lines**, becoming a thin `reader.Next()`-dispatch orchestrator; every SSE quirk is now an isolated, individually unit-tested stage.

No wire-format, metering, circuit-breaker, or logging changes — this is a pure restructuring. 17 commits, each independently green and byte-exact-verified.

## Why

Per the maintainer's review thread, real user issues are dominated by SSE-quirk reports (stray blank lines / triple `\n`, parser chokes, BOM, CR handling). Each fix used to land as another interacting boolean in the single most regression-prone function in the repo. A pipeline makes each quirk an append-only, structurally-isolated stage with its own test, so a fix for client A can no longer regress client B.

## Structure

The god-function is split into:

| Component | File | Role |
|-----------|------|------|
| orchestrator | `proxy.go` | thin `Next()` dispatch over blank / comment / `[DONE]` / data |
| `streamReader` | `stream_reader.go` | scanner (probe replay), stall watchdog, chunk count, empty-line limit, disconnect, BOM/CR cleanup, classification |
| `streamSink` | `stream_sink.go` | SSE output, byte accounting, the `\n\n`-vs-`\n` separator rule (`swallowBlank`) |
| observers | `stream_observers.go` | `observeDataChunk` (usage / native_finish / repeated-content / chunk.Error), `captureSSEError` (P1-B/P1-C) |
| transforms | `stream_transforms.go` | `computeStripReasoning`, `normalizeReasoningChunk`, `stripEmptyReasoningContent`, `computeFinishReason` |
| per-chunk pipeline | `handleDataChunk` | runs capture → parse → transforms → observers → forward; emits at most once |
| finalizer | `stream_finalize.go` | TPS, error/stall classification, missing-`[DONE]` injection, final log, CB-on-stall, metering + `streamState` |

## Tests

- **3 byte-exact golden cases** (`stream_golden_test.go`) pin the wire output: the blank-line/separator rule, the strip_reasoning keep-alive serialization, and the finish_reason rewrite. These gate every commit.
- **14 per-stage unit tests** (DB/HTTP-free) across reader, sink, both observer methods, and all four transforms.
- Full pre-existing `internal/proxy` + `internal/failover` suite passes unchanged at every commit.

## Notes for review

- The transform *compute* was extracted into tested helpers while the single emit + bookkeeping stayed inline, so each step is byte-identical (verified by the golden cases).
- Plain-forward deliberately stays `sink.write(line)` on the **raw** upstream bytes (preserves LM Studio's no-space `data:` framing) — it is *not* routed through `writeData`.
- A latent double-emit possibility between reasoning-normalize and empty-content-strip is **preserved as-is** (they are mutually exclusive in practice but not enforced); not "fixed" inside a behavior-preserving refactor.
- The commits are small and ordered (Phase 0 net → sink → finalize → reader → transforms → orchestrator); reviewable one at a time.

### DIFF
--- a/.github/workflows/crowdin.yml
+++ b/.github/workflows/crowdin.yml
@@ -27,7 +27,7 @@ jobs:
         uses: crowdin/github-action@8868a33591d21088edfc398968173a3b98d51706 # v2
         with:
           upload_sources: true
-          upload_translations: true
+          upload_translations: false
           auto_approve_imported: true
           download_translations: true
           skip_untranslated_strings: true
--- a/internal/proxy/proxy.go
+++ b/internal/proxy/proxy.go
@@ -80,12 +80,12 @@ func (h *Handler) handleStreamingResponse(w http.ResponseWriter, r *http.Request
 	// classified sseEvents; this orchestrator owns emits and transforms.
 	reader := newStreamReader(r.Context(), resp.Body, opts, logData)
 
-	var promptTokens, completionTokens, reasoningTokens int
-	var promptCacheHitTokens, promptCacheMissTokens int
-	var lastErrMsg string
-	clientDisconnected := false
-	sawDone := false
-	errorChunkCount := 0
+	// st accumulates the per-stream metrics, carry flags, and observer state
+	// (Phase 4 §6 migration). Created before the loop and mutated in place so
+	// the transforms/observers and the finalizer share one named contract
+	// instead of a fistful of loop-locals. The stall flag and final chunkCount
+	// are filled from the reader at logUpdate.
+	st := &streamState{}
 	// Periodic streaming progress logging (every 50 chunks) to give
 	// visibility into stream health without flooding logs.
 	const chunkLogInterval = 50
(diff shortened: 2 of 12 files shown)


hugalafutro:
Thanks @greptile-apps — all three addressed in 2ea88c92:

- **P1 (usage dropped on strip_reasoning early exits)** — good catch, real metering bug. It's actually **pre-existing**: the original code's strip_reasoning `continue` skipped the usage observer too; this refactor's structure just made it visible. Fixed by running `observeDataChunk` for every valid chunk **before** the transforms (which also restores the documented "observers run regardless" invariant). Observers never emit, so wire output is byte-identical — only metering/diagnostics change (now correct). Added `TestHandleStreamingResponse_StripReasoning_UsageOnFinalChunk` as a regression test (fails without the fix).
- **P2 (duplicated P1-B flush)** — extracted `(st).flushAccumulatedError`, now called from both the comment handler and `captureSSEError`.
- **P2 (missing golden)** — added `TestHandleStreamingResponse_GoldenStripReasoningKeepaliveThenComment`, pinning that the keep-alive's `swallowBlank` consumes its own separator while the comment's trailing blank is still forwarded.

Full `internal/proxy` + `internal/failover` suite green; gofmt/vet/golangci-lint clean.

PR #191, opened 2026-06-10
ci: sync Docker Hub README from tracked docs/DOCKERHUB.md on release

hugalafutro:
## What

The Docker Hub repository description was a hand-pasted copy of (parts of) the README that silently drifted from the shipped features — until today it still described chat completions as the only proxy endpoint. This PR makes it a tracked document with automatic sync:

- **`docs/DOCKERHUB.md`** — the Hub-tailored description, seeded byte-for-byte from the *current live* Docker Hub text (which already has working absolute links and the Hub-specific "Deploy without Git" section). It is deliberately a separate document from `README.md`: the GitHub README's relative links and icon paths break on Docker Hub.
- **`docker-publish.yml`** — a `Sync Docker Hub README` step (pinned `peter-evans/dockerhub-description@v5.0.0`) uploads that file verbatim on every release, reusing the existing `DOCKERHUB_USERNAME`/`DOCKERHUB_TOKEN` secrets. The PAT's read/write/delete scope was verified against the live API today. `continue-on-error: true` guarantees a failed README sync can never fail an image release.

## Notes

- Future feature docs: edit `docs/DOCKERHUB.md` alongside `README.md`; the next `.version` bump publishes both image and description together.
- No links are rewritten by the action — the file is uploaded as-is, so what renders on Docker Hub is exactly what's reviewed here.


hugalafutro:
> docs/DOCKERHUB.md — the details/summary HTML elements flagged in prior review threads are not yet resolved and will render incorrectly on Docker Hub.

Not only is this untrue as I just tested it within docker readme, but moreover it prefectly fine renders the collapsed details block as well...

PR #198, opened 2026-06-11
ci: image vuln scanning (gate + weekly SARIF), govulncheck, actionlint, run cancellation

hugalafutro:
## Why

PR #197 fixed the OpenSSL CVEs flagged on Docker Hub, but nothing prevents a recurrence: CVEs appear in already-published images with no code change, and nothing in CI would catch a regression that reintroduces vulnerable packages. Follow-up discussed on #197, expanded per discussion to carry the other agreed CI improvements in one PR.

## What

**Image vulnerability scanning** — two layers of Trivy, both gating with the same **actionable-only** policy (fail only on HIGH/CRITICAL findings that have an upstream fix, `ignore-unfixed`):

1. **PR/branch gate** (`ci.yml`, existing `docker-build` job): scans the freshly built image. `ignore-unfixed` is the make-or-break setting — without it, a new unpatched CVE upstream would block every PR with nothing anyone can do. With it, and with the `apk upgrade` from #197 pulling current Alpine patches at build time, a red gate means something real: a Go-module vuln in the binary or a Dockerfile regression.
2. **Weekly scan of published `latest`** (`image-scan.yml`, Mondays 06:17 UTC + manual dispatch): covers the gap the PR gate can't — the published image aging while no PR is open (exactly how the OpenSSL flags appeared). A failure means a rebuild/release would fix something, so it doubles as the signal to cut a patch release. It also catches the publish workflow's GHA layer cache serving a stale `apk upgrade` layer. Per review feedback, the weekly scan also uploads a **SARIF report to the Security tab** (category `published-image`) — deliberately broader than the gate (all severities, unfixed included) so findings are tracked over time.

Trivy over Docker Scout: no Docker Hub login required, so the gate also works on PRs from forks; both scanners agreed on findings in #197's verification.

**Other CI hardening** (`ci.yml`):

- **govulncheck job** (pinned v1.3.0): complements the image scan's gobinary check with reachability analysis — fails only on vulnerable functions the code actually calls.
- **actionlint job** (pinned v1.7.12): lints the workflow files, using the runner's preinstalled shellcheck for `run:` blocks.
- **Concurrency group**: cancels superseded CI runs when a PR branch is updated; pushes to master always run to completion.

## Verification

All gates run locally with the exact CI settings:

| Check | Result |
|---|---|
| Trivy on `model-hotel:ci` built from this branch (PR gate) | **0 findings, exit 0** |
| Trivy on `docker.io/hugalafutro/model-hotel:latest` (weekly scan) | **2 HIGH (OpenSSL 3.5.6, fixed in 3.5.7-r0), exit 1** |
| govulncheck `./...` | **No vulnerabilities found** |
| actionlint (all 5 workflows, incl. the ones added here) | **0 issues** |

⚠️ The published-image result means the weekly scan **will be red until the next release** is published with the #197 fix — that's the workflow doing its job (a release would clear it).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,12 @@ on:
 permissions:
   contents: read
 
+# Cancel superseded runs when a PR branch is force-pushed/updated, but never
+# cancel in-flight runs on master itself.
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   go-test:
     name: Go Test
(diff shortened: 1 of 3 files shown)


hugalafutro:
Addressed the SARIF gap in 4ad72410 — the weekly scan now uploads a SARIF report to the Security tab (category `published-image`). The report is deliberately broader than the gate: all severities including unfixed findings, so the tab tracks the full picture (e.g. the no-fix busybox medium) over time, while the gate itself still only fails on fixable HIGH/CRITICAL.

Also expanded the PR scope per discussion (ac2738da): govulncheck job (reachability-based Go vuln analysis, pinned v1.3.0), actionlint job (workflow linting, pinned v1.7.12), and a concurrency group that cancels superseded PR runs but never master runs. Both new gates verified clean locally.

PR #236, opened 2026-06-16
chore: add THIRD-PARTY-NOTICES and generator

hugalafutro:
## What

Adds a `THIRD-PARTY-NOTICES.md` bundling attribution for every distributed third-party component, plus a generator to keep it current.

## Why

Our MIT/BSD/Apache/ISC dependencies require retaining their copyright/license notices in distributions, and the SIL OFL fonts require their license files to travel with the embedded fonts. We shipped neither a notices file nor a way to produce one. This closes that gap.

## What's included

- **`tools/gen-notices/`** — collects license files from the Go modules compiled into `./cmd/server` and the production (bundled) frontend npm packages, reproducing each verbatim and de-duplicating by content. Dev-only tooling (test runners, linters, build plugins) is intentionally excluded — it isn't shipped, so it carries no attribution obligation.
- **`make notices`** — regenerates the file.
- **`.gitignore`** — tracks `tools/gen-notices/` like the existing `tools/i18n-translate/` exception.
- **`THIRD-PARTY-NOTICES.md`** — 49 Go modules + 206 npm packages, all permissively licensed (MIT / Apache-2.0 / BSD-2/3-Clause / ISC / 0BSD). Fonts (JetBrains Mono, Onest, Schibsted Grotesk, KaTeX) are SIL OFL 1.1.

## Notes

- Generated file is large (~5k lines) by nature — it reproduces full license texts. Header marks it "do not edit by hand; regenerate with `make notices`".
- No runtime/behavioral changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -287,6 +287,47 @@ jobs:
           fail_ci_if_error: false
           token: ${{ secrets.CODECOV_TOKEN }}
 
+  # Guards against THIRD-PARTY-NOTICES.md drifting from the actual dependency
+  # set: regenerate it and fail if the committed copy differs. The generator is
+  # deterministic, so any diff means someone changed deps without running
+  # `make notices`.
+  notices:
+    name: Third-Party Notices
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+
+      - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6
+        with:
+          go-version-file: go.mod
+          cache: true
+
+      - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6
+        with:
+          version: 10
+
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
+        with:
+          node-version: 24
+          cache: pnpm
+          cache-dependency-path: web/pnpm-lock.yaml
+
+      # make notices shells out to `pnpm licenses list`, which needs the
+      # production graph installed from the committed lockfile.
+      - name: Install frontend dependencies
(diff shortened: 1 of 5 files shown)


hugalafutro:
## Review follow-up: fixes + two won't-fix decisions

Addressed the substantive findings; documenting the two I'm intentionally not changing, with evidence.

### Fixed
- **Stale versions (P1)** — regenerated against the committed lockfile; dropped the generated-on date so output is reproducible.
- **Multiple-versions collapse (P2)** — `collectNPM` now emits every version/path pair instead of pairing one version with another's license path.
- **Extra license files skipped (P2)** — `readLicenseFile` collects all license/copying files (dual-licensed packages), deduped + deterministic.
- **Multiple NOTICE files collapse (P1)** — same treatment now applied to `NOTICE`/`NOTICE.*` files (was last-wins).
- **Unpinned pnpm — stderr half** — command failures now surface `stderr`.
- Added a CI job that runs `make notices` and fails on any diff, so the file can't silently drift.

### Won't-fix #1 — "Dev packages included" (`@types/react`, `typescript`)
The premise (these are dev-only) is incorrect: both are in the **production** dependency graph, pulled transitively by runtime deps. `pnpm why --prod`:

```
typescript@6.0.3
└─┬ i18next@26.3.1            (runtime dependency)

@types/react@19.2.17
├─┬ react-markdown@10.1.0     (runtime dependency)
└─┬ react-redux → recharts@3.8.1   (runtime dependency)
```

`pnpm licenses list --prod` is therefore correctly including them. Attributing the full production dependency closure is the intended, defensible behavior — narrowing it to "bytes in the final bundle" would under-attribute. No change.

### Won't-fix #2 — "Pin pnpm via Corepack"
Reproducibility is already covered without adding a Corepack dependency to a local dev tool:
- CI installs the pinned pnpm via `pnpm/action-setup` before running `make notices`.
- The new drift-check job regenerates and fails on any diff, so a divergent CLI would be caught immediately.

The actionable half of that finding (surface `stderr`) is done.

PR #287, opened 2026-06-22
fix(provider): detect minimaxi.com as Anthropic-compatible endpoint

2233admin:
## What

api.minimaxi.com/anthropic is MiniMax's official Anthropic-compatible API for their Coding Plan. `DetectProviderType` only matched `.anthropic.com`, causing minimaxi to fall back to OpenAI discovery which returns 404.

Also includes a CRLF fix for docker-entrypoint.sh during build (Windows compatibility).

## Commits

1. `fix(provider): detect minimaxi.com as Anthropic-compatible endpoint` — main fix
2. `fix(docker): strip CRLF from docker-entrypoint.sh during build` — Windows build fix

## Testing

Verified on NAS deployment: MiniMax Coding Plan provider now correctly uses Anthropic-compatible model discovery.

Closes: (add issue number if applicable)

### DIFF
--- a/internal/provider/discovery.go
+++ b/internal/provider/discovery.go
@@ -160,6 +160,9 @@ func DetectProviderType(baseURL string) string {
 	if strings.HasSuffix(host, ".anthropic.com") {
 		return "anthropic"
 	}
+	if strings.HasSuffix(host, ".minimaxi.com") {
+			return "anthropic"
+	}
 	if strings.HasSuffix(host, ".x.ai") {
 		return "xai"
 	}


hugalafutro:
I can't find the endpoint you propose mentioned anywhere in official minimax documentation, minimaxi.com (with an I at the end) appears to be a Chinese reseller from what I can gather from the translation. If you either point me in the direction of where this endpoint is documented in minimax official documentation, or amend the PR to concern the official endpoint I will take another look at it.

hugalafutro:
Thanks for the detail, and apologies, you're right that minimaxi.com is legit. I had it backwards: minimaxi.com is MiniMax's official China-mainland host and minimax.io is the global one, both first-party. So no objection there.

The thing I want to flag is that I think the PR is fixing the smaller half of the problem. There are two separate things going on:

- Model discovery (the one-time "list this provider's models" call). Your change makes this work because MiniMax's /anthropic path exposes the model list in Anthropic's shape (GET /v1/models, x-api-key auth), so the app's anthropic discoverer fits it. That's what you saw populate.
- Proxying (every actual chat request). This is the part the PR doesn't touch, and it's where I think it'll bite. Model Hotel only ever speaks OpenAI format upstream. It has no native Anthropic /v1/messages client anywhere. The provider type we call anthropic doesn't mean "talk Anthropic Messages", it means "send an OpenAI-shaped body to <base>/v1/chat/completions with x-api-key". That works against real Anthropic only because they run an OpenAI-compat shim at that path.

So with your base URL, an actual completion would get POSTed as OpenAI JSON to https://api.minimaxi.com/anthropic/v1/chat/completions, and per MiniMax's own docs the /anthropic path serves native Anthropic Messages (/anthropic/v1/messages), not OpenAI chat/completions. Discovery lights up, but I'd expect the first real request to 404 or error. Did you manage to run an actual chat through it, or just see the models appear?

The simpler route that should just work today, no code change: add MiniMax as a normal custom provider pointed at their OpenAI-compatible endpoint, https://api.minimaxi.com/v1, and leave the type on auto-detect (it'll be treated as openai). That's the format the proxy actually speaks. If discovery there 404s on /v1/models, that's the real narrow bug and I'd happily take a fix for that instead.

(You may also need to allow the host via ALLOWED_PROVIDER_HOSTS if it isn't already.)

Let me know if that helps. Bottom line is even if I added the host to where the pr wants it it would only make model discovery work and the requests would be pointed at Anthropic shaped endpoint not /v1 openai endpoint which is (for now and foreseeable future) the only type of endpoint Model Hotel supports

hugalafutro:
Feel free to open a new pr or an issue if the proposed solution did not work for you, again MH only proxies /v1 OpenAI data, any and all references to anthropic in the code to date is dealing with their openai /v1 compatibility endpoints.

PR #297, opened 2026-06-25
feat(ha): fleet config sync (Phase 5, v1)

hugalafutro:
## What

HA Phase 5: replicate one member's configuration to the rest of the fleet from **Front Desk → Settings → Config sync**, so a fresh member that started empty ends up identical to the chosen primary without hand-copying provider keys.

Pick a **primary** (config source of truth); Front Desk pulls its config and pushes it to every other member. This is a **separate action** from admin-token sync, by design: a config replace can remove providers/keys on a replica, so it must not ride along with a routine token rotation.

## Design decisions

- **JSON over the member admin API, not the pg_dump backup.** The backup is the whole DB (logs, metering, events) and destructive to restore; wrong tool for config. New endpoints `GET /api/config/export` and `POST /api/config/import?dryRun=`.
- **Shared `MASTER_KEY` ports the encrypted provider keys verbatim.** A replica decrypts them with its own copy; nothing is re-entered, and Front Desk never sees plaintext.
- **Cross-instance refs travel by name, never instance-local UUID.** A virtual key's `allowed_providers` stores provider UUIDs (the proxy compares `c.provider.ID.String()`), so it is exported as provider names and re-resolved to each importing member's own UUIDs. This is exactly why v1 scopes to the entities without other local-ID refs.
- **Declarative upsert, primary wins, never silent.** A dry-run diff (added / updated / removed per entity) is shown and acknowledged before any write.
- **No data loss on replace.** Providers upsert in place; removing a provider cascades only its discovered models, while `request_logs` survive (`provider_id` FK is `ON DELETE SET NULL`, migration 010) and virtual keys carry no log FK. There is a test asserting exactly this.

## Scope

**v1 (this PR):** providers (incl. encrypted keys), virtual keys (by hash), and the syncable settings subset (`settings.AllowedSettings` minus the per-instance apprise destination). One predicate (`isSyncableSetting`) gates export, diff, and apply.

**Deferred (documented):**
- **Models and failover groups are not copied** — each member regenerates them from the synced providers (discovery + automatic group formation). Their manual overrides (disable / rename / custom priority) need model-ID translation and are a follow-up.
- **Models appear after discovery, not instantly.** Import does not yet trigger a discovery pass, so a freshly synced replica shows its providers immediately but its models populate on the next discovery cycle (or a manual Discover). A post-import discovery trigger is the natural next enhancement.
- Alerting destination stays per-instance; continuous/auto sync and per-entity selection are future.

## Guards

- MASTER_KEY canary decrypt → **409**, writes nothing (a member that can't use the keys is flagged "blocked" in the preview).
- Unknown `schema_version` → **422**; structurally empty envelope → **400** (refuses to wipe a target).
- Import runs in a single transaction (all-or-nothing per member); caches invalidated after apply.
- Diffs and responses carry names/counts only, never key material; nothing logged.

## Tests

- Member side (`internal/api/configsync_test.go`): export/import round trip incl. `allowed_providers` UUID↔name translation, MASTER_KEY mismatch, dry-run-writes-nothing, declarative-replace-preserves-logs-and-models, empty/ wrong-schema refusals.
- Front Desk (`internal/frontdesk/configsync_test.go`): preview classification (matches / overwrite / blocked / mismatch), apply, failure reporting + events.
- UI (`ConfigSyncPanel.test.tsx`): preview diff + removal warning + double-confirm; MASTER_KEY-mismatch blocked path.
- `go build ./...`, full `internal/api` + `internal/frontdesk` suites, `go vet`, golangci-lint clean; FD web `tsc -b`, `typecheck:tests`, eslint, biome clean, vitest 47/47.

## Reviewer notes

- The Front Desk UI is embedded in the binary; a rebuild (`make ha-up` / `make frontdesk-build`) is needed to see the panel.
- Docs: `docs/HA.md` + `wiki/High-Availability.md` "Replicating config".


hugalafutro:
Addressed the review in `04016b8c`:

- **P1 (check order)** — `previewMemberConfig` / `applyMemberConfig` now test `SchemaVersionOK` before `MasterKeyOK`. A 422 schema rejection happens before the member runs the MASTER_KEY canary, so `master_key_ok` was an unevaluated false and a version skew was misreported as a key mismatch. Added tests for the schema-mismatch diagnosis (the old stubs masked it by setting `master_key_ok:true` on a 422).
- **P2 (settings not deleted)** — settings are now declaratively replaced like providers/VKs: `computeDiff` reports `Settings.Removed` and `apply` deletes syncable keys the primary lacks (replica falls back to the same built-in default), never touching non-syncable keys (apprise / observability / instance-local). Cache invalidation + `NotifyDeleted` for removed keys moved into `apply`.
- **P2 (stale `allowed_providers` widening)** — a virtual key whose restriction resolves to zero providers on the importing member is now skipped rather than written with an empty allow-list (which the proxy treats as "all allowed"). Export comment updated. In the normal flow this never triggers since providers upsert before VKs in the same transaction.

Also raised patch coverage (new-file statement coverage now ~84% api / ~88% frontdesk): tests for diff add/update/remove, keyless import, settings removal + non-syncable preservation, the VK-skip path, and handler decode/DB-error branches.

PR #315, opened 2026-06-28
feat: configurable session inactivity auto-logout (dashboard + Front Desk)

hugalafutro:
## What

Signs the admin/operator out after a configurable period of inactivity, default 60 minutes; `0` disables it. Built for both frontends (Model Hotel dashboard and Front Desk).

## How

**Shared helper** (`utils/idleLogout.ts`, an identical copy in `web/` and `frontdesk/web/` since they are separate Vite roots that cannot import across each other): a dependency-free idle timer with activity listeners, cross-tab deadline coordination via a `localStorage` timestamp + `storage` event, and a `visibilitychange` re-check for suspended tabs. Each app wraps it in a thin `useIdleLogout` hook that reuses the app's existing logout path. A test byte-compares the two copies and fails on drift.

**Model Hotel**: new "Session timeout" card in the Authentication settings pill (slider, `∞ = Never`). Stored as `session_idle_timeout_minutes` in the settings K/V bag (added to both allowlists, no migration) and kept instance-local so config sync never replaces it and managed members can still edit it. English strings translated into all 28 locales.

**Front Desk**: `session_idle_timeout_minutes` on the `Settings` struct + migration `008`, a number field in Settings, wired into the Shell.

**On timeout**: silent logout (clear token, best-effort server-side revoke, back to login).

## Security: server-side revoke on every auth path

`handleLogout` no longer gates the revoke behind WebAuthn availability, so a TOTP-only admin's session is revoked server-side too (the endpoint deletes whatever session matches the bearer; a raw token with no session is a no-op). Front Desk gained `POST /api/logout` backed by its existing `SessionManager`.

## Tests / verification

- Go: handler round-trip test per the AGENTS.md allowed-setting rule, FD store default/round-trip/bounds, `TestServerLogout`, allowlist-sync, configsync/fleet.
- TS: `tsc -b` both apps; new idle-logout suites (throttle, visibilitychange-past-deadline, seed-from-peer, cross-root drift) + Layout/Settings/App suites.
- Bounds aligned to `0..240` across MH validator, FD validator, and the FD number field (no more silent clamp / overflowing slider fill).
- `golangci-lint`, `biome`, `eslint` clean; i18n parity check passes for all 28 locales.


hugalafutro:
Thanks, addressed in c8e8ea06.

**P1 (Stale deadline survives logout) — fixed.** Real bug. On startup `startIdleLogout` now seeds from a stored timestamp only when its deadline is still in the future; a stale value left behind by a logged-out session falls through to a fresh window (`recordActivity`), which also overwrites the lingering key. So a fresh login no longer arms a zero-delay timer and bounces back to login. Done self-contained in the helper so both byte-identical copies stay in sync (the `idleLogout.identical` drift test enforces it), with a regression test added to each suite.

**P2 (Null timeout disables logout) — does not reproduce.** The chain breaks at "Go decodes the int field to 0". `encoding/json` treats a JSON `null` into a non-pointer `int` as a no-op, so the partial-merge decode (`GetSettings` -> decode body on top) preserves the stored value, it does not zero it. Verified empirically (null into `int{60}` -> 60) and added a regression test that PUTs `{"session_idle_timeout_minutes": null}` and asserts the stored value is unchanged. `0` remains a deliberate, validated "never" that only an explicit `0` can set.

PR #338, opened 2026-07-01
fix(ha): tolerate rebuild churn in the HA plane

hugalafutro:
## Why

Rebuilding an HA member (or several containers at once) to `:dev` produced ~10-15 orange **app** error-shelf entries like:

```
POST host=model-hotel1.zmrd.uk path=/api/fleet/announce status=500 duration=3.99944149s
```

Three separate issues, each fixed in its own commit.

## What

**1. `fix(errorshelf)` — categorize `/api/fleet/*` access-log 5xx as HA, not app**
The shelf colors by slog *source*; the HTTP access logger stamps every request line source `access`, so a fleet control-plane 5xx missed the indigo HA chip (which keyed only on `configsync`/`fleet` sources). `isHaAccessLog(source, message)` recovers the HA classification from the `path=/api/fleet/` field. Frontend-only.

**2. `fix(fleet)` — record the announce heartbeat in one DB round-trip**
The `3.99944149s` was Front Desk's 4s announce-client timeout. The member's `Announce` handler did four sequential `settings.Set` writes; under rebuild DB contention those four round-trips blew the budget, the client cancelled, and the member logged a spurious 500. New `settings.Repository.SetMany` does a single multi-row upsert (same cache-evict + subscriber-notify semantics as `Set`), all-or-nothing. Added to the `SettingsStore` + `fleetSettings` interfaces and both test mocks.

**3. `feat(frontdesk)` — debounce member health-down with a configurable threshold**
`applyHealth` flipped a member down on the *first* failed poll, emitting a `health.down` error event and a default-on Apprise alert — so every routine rebuild raised an alert the instant the container blinked. Now a member must miss `health_fail_threshold` polls in a row before it is reported down (mirrors the existing `versionFetchFailThreshold` debounce). Recovery stays immediate; the last known-good badge is kept during the grace window so the dashboard no longer flickers red. New Front Desk setting (migration 011, default 3 ≈ 15s at the 5s poll interval), tunable on the Settings page.

**4. `feat(frontdesk)` — damp the Traefik UP→DOWN badge flip**
Traefik briefly stops listing a member during a restart; committing that immediately flapped the badge and re-nudged every dashboard. A non-UP status is now held back until observed `health_fail_threshold` polls in a row (same knob); UP still applies at once, so recovery is immediate.

## Testing
- Backend: full `internal/frontdesk` suite + targeted `internal/settings` (`SetMany`) and `internal/api` (fleet) tests pass; `golangci-lint` clean on all touched packages.
- Frontend: `tsc -b` clean on `web/` and `frontdesk/web/`; ErrorShelf, SettingsPage, AlertsPanel, OidcPanel suites pass; biome clean.
- New tests cover: batched announce write + all-or-nothing failure, `SetMany` upsert + subscriber notify, health debounce (grace-window silence, threshold crossing, immediate recovery, configurable threshold, first-observation-down), and Traefik down-flip damping.

## Notes
- If merged: needs a `.version` bump + dev rebuild (the FD Settings field is embedded, so `docker compose restart` won't pick it up).
- Operators who routinely take 30s+ to rebuild can raise `health_fail_threshold` on the FD Settings page rather than tolerate an alert.

### DIFF
--- a/frontdesk/web/src/api/types.ts
+++ b/frontdesk/web/src/api/types.ts
@@ -54,6 +54,10 @@ export interface Settings {
 	traefik_stale_secs: number;
 	event_retention_days: number;
 	retry_attempts: number;
+	// Consecutive failed health polls before a member is reported down (an error
+	// event plus, by default, an Apprise alert). Damps routine-rebuild flap; also
+	// governs the Traefik UP->DOWN badge flip. Minimum 1.
+	health_fail_threshold: number;
 	// Admin-UI inactivity auto-logout window in minutes; 0 disables it. Consumed
 	// by useIdleLogout to sign the operator out after inactivity.
 	session_idle_timeout_minutes: number;
--- a/frontdesk/web/src/components/__tests__/AlertsPanel.test.tsx
+++ b/frontdesk/web/src/components/__tests__/AlertsPanel.test.tsx
@@ -34,6 +34,7 @@ const settings: Settings = {
 	traefik_stale_secs: 30,
 	event_retention_days: 90,
 	retry_attempts: 2,
+	health_fail_threshold: 3,
 	session_idle_timeout_minutes: 60,
 	alert_enabled: true,
 	alert_apprise_api_url: "http://apprise:8000",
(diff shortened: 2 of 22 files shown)


hugalafutro:
Addressed both review comments in a1aaa985:

- **poller.go badge nudge** — `applyHealth` now compares both `Known` and `Healthy` when deciding whether to publish a badge nudge, so a never-seen member crossing straight from "unknown" to confirmed-down (Known flips, Healthy doesn't) still refreshes the badge, not just the event log. Covered by an extended `TestApplyHealthFirstObservationDownDebounced`.
- **settings.SetMany notify** — documented that the row write is atomic (one statement) but the post-commit change notifications are per-key, fanned out like a loop of `Set` calls, so a subscriber may see the keys arrive one at a time.

PR #353, opened 2026-07-04
fix(web): hide non-chat modalities from Chat and Arena pickers

hugalafutro:
## What

Embedding and rerank models were leaking into the **Chat** and **Arena** model pickers, where they can never work: selecting one and pressing send routes a chat completion to a model that only serves `/v1/embeddings` or `/v1/rerank`. This surfaced during the rerank PR (rerank newly leaked in; embedding leakage was pre-existing) and was deliberately left for this focused follow-up.

## How

- **`utils/model.ts`** — new `isChatModel()` predicate + `NON_CHAT_MODALITIES` set (`embedding`, `rerank`, `image`, `tts`, `stt`). Denylist, default-allow: an unknown or empty modality is treated as chat so a future modality never silently disappears from the picker.
- **`hooks/useModels.ts`** — new `useChatModels()` layered on the existing `useEnabledModels()`.
- **`Chat/useChat.ts`** and **`Arena/useArenaState.ts`** — source their models from `useChatModels()`. The random-model actions filter this same list, so they're covered automatically.

`ModelPicker.tsx` itself is unchanged: it's shared with the failover group editor, which must see every model, so the filter lives at the chat/arena caller.

## Intentional parity notes

1. **`/v1/models` and the failover group editor still list every model by design** — the filter is picker-side only. OpenAI-compatible clients reach embedding/rerank models through `/v1/embeddings` and `/v1/rerank`.
2. **Known gap, not addressed here:** locally-discovered embedding models (LM Studio and other OpenAI-compatible locals) are stored with `modality:"text"` in backend discovery, so a modality-based filter can't catch them. That's a pre-existing backend classification issue and a separate follow-up.

## Tests

- `isChatModel` unit tests (chat modalities → true; `embedding`/`rerank`/`image`/`tts`/`stt` → false; case-insensitive; default-allow unknown/empty/missing).
- `useChatModels` drops embedding/rerank rows, keeps chat rows.
- Chat/Arena hook-mock exports updated to the new hook name.

`tsc -b` clean, full Chat + Arena suites green, Biome/ESLint clean on changed files. No i18n changes (pure filtering), no backend changes.

### DIFF
--- a/web/src/hooks/__tests__/useModels.test.tsx
+++ b/web/src/hooks/__tests__/useModels.test.tsx
@@ -6,6 +6,7 @@ import type { Model, Provider } from "../../api/types";
 import { mockModel, mockProvider } from "../../test/mocks/data";
 import { server } from "../../test/mocks/server";
 import {
+	useChatModels,
 	useEnabledModels,
 	useModels,
 	useProviderData,
@@ -189,6 +190,42 @@ describe("useEnabledModels", () => {
 
 		expect(result.current.data).toHaveLength(1);
 		expect(result.current.data?.[0].provider_name).toBe(mockProvider.name);
+	});
+});
+
+describe("useChatModels", () => {
+	it("excludes non-chat modalities (embedding, rerank)", async () => {
+		const embeddingModel: Model = {
+			...mockModel,
+			id: "model-embedding",
+			model_id: "text-embedding-v1",
+			modality: "embedding",
+		};
+		const rerankModel: Model = {
+			...mockModel,
+			id: "model-rerank",
+			model_id: "rerank-v1",
+			modality: "rerank",
+		};
+
+		server.use(
+			http.get("/api/models", () =>
+				HttpResponse.json([mockModel, embeddingModel, rerankModel], {
+					status: 200,
+				}),
+			),
+		);
+
+		const { result } = renderHook(() => useChatModels(), {
+			wrapper: createWrapper(),
+		});
+
+		await waitFor(() => {
+			expect(result.current.isSuccess).toBe(true);
+		});
+
+		expect(result.current.data).toHaveLength(1);
+		expect(result.current.data?.[0].id).toBe(mockModel.id);
 	});
 });
 
--- a/web/src/hooks/useModels.ts
+++ b/web/src/hooks/useModels.ts
@@ -2,6 +2,7 @@ import { useQuery } from "@tanstack/react-query";
 import { useMemo } from "react";
 import { api } from "../api/client";
 import type { Model } from "../api/types";
+import { isChatModel } from "../utils/model";
 
 /**
  * Fetch all models. React Query deduplicates by queryKey - multiple
@@ -28,7 +29,8 @@ export function useProviders() {
 
 /**
  * Enabled models only - filters to models that are both enabled and
- * have a provider assigned. Used by Chat and Arena for model pickers.
+ * have a provider assigned. Base list for the chat surfaces, which layer
+ * useChatModels on top to also drop non-chat modalities.
  */
 export function useEnabledModels() {
 	const { data: models, ...rest } = useModels();
@@ -37,6 +39,18 @@ export function useEnabledModels() {
 		[models],
 	);
 	return { ...rest, data: enabledModels };
+}
+
+/**
+ * Enabled, chat-capable models. Excludes embedding/rerank (and other non-chat)
+ * modalities so they don't appear in the Chat and Arena pickers, where they
+ * could never be used. The failover group editor and /v1/models keep listing
+ * every model.
+ */
+export function useChatModels() {
+	const { data: models, ...rest } = useEnabledModels();
+	const chatModels = useMemo(() => models.filter(isChatModel), [models]);
+	return { ...rest, data: chatModels };
 }
 
 /**
--- a/web/src/i18n/locales/af.json
+++ b/web/src/i18n/locales/af.json
@@ -2651,7 +2651,8 @@
 		"useArenaRunner": {
 			"generationError": "{{model}}: {{error}}",
 			"retry": "{{model}} {{status}} – probeer weer {{attempt}} in {{delay}}s…",
-			"networkError": "netwerkfout"
+			"networkError": "netwerkfout",
+			"nonChatModel": "Nie 'n geselsmodel nie"
 		},
 		"useQuotaData": {
 			"nanoGPTError": "Misluk om die NanoGPT-gebruikkwota op te haal",
--- a/web/src/i18n/locales/ar.json
+++ b/web/src/i18n/locales/ar.json
@@ -2651,7 +2651,8 @@
 		"useArenaRunner": {
 			"generationError": "{{model}}: {{error}}",
 			"retry": "{{model}}: {{status}} - إعادة المحاولة {{attempt}} في {{delay}}s…",
-			"networkError": "خطأ في الشبكة"
+			"networkError": "خطأ في الشبكة",
+			"nonChatModel": "ليس نموذج محادثة"
 		},
 		"useQuotaData": {
 			"nanoGPTError": "فشل في جلب حصة استخدام NanoGPT",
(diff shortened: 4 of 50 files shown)


hugalafutro:
Addressed the Greptile finding (persisted selections can remain outside the filtered list) in df6c6c9e.

Both chat surfaces now reconcile persisted selections against the chat model list once it loads:
- **Chat** clears `chatSelectedModel` / `conversationModelA` / `selectedModelB` if the persisted id isn't a valid chat model.
- **Arena** filters the `bracket`/`compare` arrays, **setup phase only** so a running competition's captured line-up is never mutated.

Both guard on a non-empty list so a transient empty fetch can't wipe a valid selection. This covers the realistic triggers (a model reclassified to embedding/rerank, or disabled) without needing an unavailable-state UI path.

Also realigned the Arena hook tests to production proxy-id format (`provider/model`) so the new validity check sees valid ids, and added drop-stale-selection coverage for both Chat and Arena. Full frontend suite green (5279 passed).

hugalafutro:
Second review round (2 P1s) addressed in bc11215d.

**Chat regenerate sent empty model id** (`useChat.ts:193`): `handleRegenerate` ran `selectedModel || ""`, so a just-reconciled (cleared) stale selection became a chat request with an empty model. Added the same `!selectedModel` guard `handleSend` already has.

**Persisted Arena competition streamed a stale slot model** (`useArenaRunner.ts:355`): a competition reloaded outside the setup phase (where array reconciliation is skipped) could keep a round slot pointing at a now-invalid model and stream it on run/retry/next-round. Guarded the runner's single dispatch choke point (`streamModel`): if the slot model isn't in the chat-filtered `enabledModels`, the slot is stamped as an errored response instead of hitting the non-chat endpoint. The guard is skipped while the list is still loading so a transient empty fetch never errors a valid competition.

Added `hooks.useArenaRunner.nonChatModel` across all 28 locales (i18n parity check green), realigned the runner tests to production proxy-id format, and added coverage for both guards. Full frontend suite green (5281 passed).

hugalafutro:
Third review round (1 P1) + codecov patch gaps addressed in 974f86ba.

**Empty allowlist skipped validation** (`useArenaRunner.ts:152`): the `validModelIds.size > 0` escape hatch I added last round left the bypass open during the model-list load window, so a persisted non-chat slot could still reach `/api/chat/arena`. Dropped it: the guard now always blocks a model absent from the chat-filtered list. Runs are user-initiated after the page (and its app-wide-cached model list) has rendered, so an unrecognised slot id is a genuine non-chat model, not a load race; blocking is the safe direction. The T-Rex repro scenario (empty list) is now covered by a test asserting the slot is stamped errored rather than dispatched.

**Codecov patch gaps** closed on the same push:
- Chat: the `conversationModelA` / `selectedModelB` reconciliation branches and the regenerate no-model guard.
- Arena runner: the guard's running-set cleanup (delete + empty-set phase flip for both compare→finished and competition→voting), slots A and B, the empty-list block, and the defensive out-of-range-matchup / non-empty-running-set edges.

Local coverage confirms 0 uncovered/partial changed lines in both files. Full frontend suite green (5286 passed).

hugalafutro:
Fourth review round (2 P1s) addressed in 9b0b6276. These were the two sides of the model-list load window:

The fix threads a `modelsReady` flag (from the chat-models query's loading state) so dispatch **defers until the list has settled** instead of guessing:

- **Chat** (`useChat.ts:185`): `handleSend`, `handleRegenerate`, and conversation start now wait for `modelsReady` before dispatching, so a persisted stale (now non-chat) selection can't be sent during the pre-reconciliation window.
- **Arena runner** (`useArenaRunner.ts:143`): while the list is still loading, an unrecognised slot model is **cleared (retryable)** rather than permanently stamped `Not a chat model`; once the list has loaded, a genuinely non-chat model is stamped errored as before.

So loading → defer (no dispatch, no permanent failure); loaded → enforce (non-chat blocked). No new i18n. Added coverage for both deferral paths; local diff-intersected coverage shows 0 uncovered/partial changed lines across the four touched files. Full frontend suite green (5290 passed).

hugalafutro:
Fifth round (2 P1s) addressed in e12ef80e. This was the loaded-*empty* edge: the models query finished with zero chat rows (or a transient error), so `modelsReady` is true but the allowlist is empty.

Unifying principle: **an empty/failed allowlist is not authoritative**, so never enforce against it.

- **Chat** (`useChat.ts:192`): reconciliation now runs once the list has *settled* (gated on `modelsReady`, not list length), so a loaded-but-empty or failed list clears every stale selection and send/regenerate/conversation can't dispatch it. The loading window stays covered by the existing `modelsReady` action guards.
- **Arena runner** (`useArenaRunner.ts:175`): the permanent non-chat error is only stamped when there's a *usable* allowlist (loaded AND non-empty); an empty list clears the slot (retryable) instead of failing valid saved-competition models.

Also seeded the send/regenerate tests' selected model ids so reconciliation leaves them intact, and added coverage for the loaded-empty clear (Chat) and defer (Arena) paths. Diff-intersected coverage: 0 uncovered/partial changed lines. Full suite green (5291 passed).

hugalafutro:
Sixth round (1 P1) addressed in 6f3ec13b.

The previous defer path (empty allowlist → clear the pending response + remove the model from `runningModels`) advanced the phase to voting/finished when it was the last running slot, erasing a valid persisted round.

Fix: stop starting or mutating an Arena run without a usable allowlist (loaded AND non-empty).
- The run initiators (`runRound`, `handleRetry`, `handleSwapComplete`) now return early when there's no usable allowlist, so a run never begins against an empty/loading list and no state is set to running/pending.
- `streamModel`'s unusable-allowlist branch is now a pure no-op (no response/running/phase mutation), so even a directly-invoked slot can't erase the round.

State is left untouched and retryable once a real model list arrives. Added coverage for all three initiator gates plus the streamModel no-op, and updated the defer tests to assert the slot stays pending rather than cleared. Diff-intersected coverage: 0 uncovered/partial changed lines. Full suite green (5294 passed).

hugalafutro:
Seventh round (2 P1s, same root cause) addressed in 2198ea5d.

Both were the missing recovery side of the run gates: a run can be marked `running` (initial dispatch, or a vote-advanced round) while the model list is still loading/empty, so every slot defers and nothing streams, leaving slots stuck at `done:false` forever once the list later becomes usable.

Fix: a recovery effect in `useArena` that, on the not-usable → usable transition of the allowlist, re-dispatches the current round when the phase is still `running`, nothing is actually streaming (abort map empty), and a slot is still pending. It fires only on the transition (not on every render while usable) so it never races the normal staggered dispatch, which already runs with a usable allowlist.

So: unusable allowlist → the run defers untouched; list becomes usable → the deferred round resumes automatically. Added coverage for the resume path plus the four guard branches (streaming in flight, non-running phase, missing round, all slots done). Diff-intersected coverage: 0 uncovered/partial changed lines. Full suite green (5300 passed).

hugalafutro:
Eighth round (1 P1) addressed in 3be2b9d1.

The recovery effect only resumed on the not-usable → usable *transition*, so a list that settles genuinely empty (zero chat-capable models) kept `hasUsableAllowlist` false forever: a saved competition reloaded in `running` with pending slots stayed stuck at `done:false` with no active stream and no exit.

Fix: extend the recovery effect with a terminal branch. Once the list has settled (`modelsReady`) with no chat models, a still-`running` round with pending slots is dropped back to `setup` instead of wedging. That branch is condition-gated (an empty list can already be settled on mount, with no transition to observe) and self-limits once the phase leaves `running`; the re-dispatch branch stays transition-gated so it never races the normal staggered dispatch.

So the full lifecycle is now closed: loading → defer; settled+models → resume; settled+empty → back to setup. Added coverage for the setup fallback. Diff-intersected coverage: 0 uncovered/partial changed lines. Full suite green (5301 passed).

hugalafutro:
Ninth round (1 P1) addressed in ac67886b.

The recovery effect only re-dispatched on a false→true allowlist transition. On a warm model cache the allowlist is usable on the very first render, so there's no transition to observe, and a saved `running` round reloaded with `done:false` slots and no active stream stayed stuck.

Fix: initialize the "previous usable" ref to `false` instead of the current value, so the first eligible render counts as a transition into usable and recovers the pending round. A normal in-session run start happens after mount (ref already `true` by then), so it's still never double-dispatched; the phase guard also blocks any mount-time action unless the phase is already `running`. One-line change, big coverage.

Added a warm-cache mount-recovery test. Diff-intersected coverage: 0 uncovered/partial changed lines. Full suite green (5302 passed).

PR #399, opened 2026-07-10
fix(discovery): dedupe providers in the Unchanged summary section

hugalafutro:
## What

The background discovery-changes modal builds one summary entry per stored change *row* (`Layout.tsx`), so a single provider can accumulate several entries across background runs: e.g. one run records a price change, another records a model removal. The **Unchanged** section keyed its chips by the unique per-row key, so once those entries resolved to no net change the same provider name rendered as two identical chips ("Provider A, Provider A").

Retesting is what surfaces it: confirming a removal that already happened turns that entry's diff empty, dropping it into Unchanged next to the sibling row.

## Fix

- Dedupe the Unchanged list by provider name.
- Drop any provider from Unchanged that already appears in the Changed section above, so a name shows at most once and never in both places.

## Tests

Added two cases to `DiscoverySummaryModal.test.tsx`:
- an unchanged provider with multiple entries lists once
- a provider with a real change is omitted from Unchanged

File suite green (12/12), `tsc -b` clean, biome clean.

### DIFF
--- a/web/src/pages/Providers/DiscoverySummaryModal.tsx
+++ b/web/src/pages/Providers/DiscoverySummaryModal.tsx
@@ -430,12 +430,17 @@ export function DiscoverySummaryModal({
 
 	const showHeaders = results.length > 1;
 	const visible = results.filter((r) => !entryIsUnchanged(r));
-	// The Unchanged section is a flat name cloud, so it dedupes by provider name:
-	// the same provider can produce several summary entries (e.g. separate
-	// background runs recorded for a price change and a model change), and once
-	// they resolve to no net change they'd otherwise list the same name twice.
-	// Providers already shown above as Changed are dropped too, so a name never
-	// appears in both places at once.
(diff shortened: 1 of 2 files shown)


hugalafutro:
Re: the failover-fallback finding (confidence 4/5) — this is intended behavior, now documented and pinned by a test in `55256297`.

The Unchanged section is a flat **name cloud**: chips are labels only, and unchanged entries have empty diffs by definition, so they carry no per-entry detail. Two entries that render the identical "Failover" chip are therefore indistinguishable, and collapsing them removes exactly the duplicate-label noise this PR targets (the same reason we collapse a real provider that appears twice). Hiding an unchanged "Failover" when a changed "Failover" is already shown above loses nothing actionable, since the changed entry renders its full diff and the unchanged one is empty.

Added `collapses failover-fallback entries sharing the same label` to lock this in, plus a clarifying code comment.

PR #408, opened 2026-07-11
feat(frontdesk): Bellhop Phase F0 device pairing

hugalafutro:
## Summary

Front Desk backend + UI prerequisites for the **Bellhop** Android companion app (Phase F0). Adds device pairing, revocable device tokens, and role-tiered API access, so a phone can be linked to a Front Desk and receive push alerts.

## Backend
- Migration `014_paired_devices` (SHA-256 hashed tokens, soft-revoke)
- Store CRUD (`store_devices.go`)
- In-memory one-time pairing codes (3 min TTL, cap 20, voided on restart)
- Admin-only `POST /api/pair/start`; public IP-limited `POST /api/pair` (code → device token, returned once)
- `requireAuth` accepts device tokens (stamps `last_seen`)
- Three route tiers: reads (monitor+), `requireOperator`, `requireAdmin`
- `DELETE /api/devices/self` (Bellhop unlink); admin `GET/DELETE /api/devices[/{id}]`

## Front Desk web
- `PairedDevicesPanel`: QR + copyable pairing string, live list poll while a code is active, revoke modal
- ntfy phone-topic helper in `AlertsPanel` (`utils/ntfy.ts`)
- PWA manifest
- 32 new keys across all 11 FD locales

## Verification
Green: Go tests (5330), frontdesk pkg (204), FD web tests (154), golangci-lint, `tsc` (app + tests), i18n-check.


hugalafutro:
Thanks Greptile. Addressed in 31582ff0:

**P1 — Device lookup blocks admin auth** (`requireAuth`): fixed. A device-token lookup error (missing/corrupt table, transient DB blip) now falls through to the admin/session gate instead of returning 500. The admin/session path never reads `paired_devices`, so a broken device table can no longer take down the whole control plane; a genuine device token simply fails auth and the phone retries. Test updated to assert the fall-through (401).

**P2 — Pairing code burns too early** (`handlePair`): fixed. `consume` now returns the full code entry and the code is `restore`d if token minting or `CreatePairedDevice` fails, so a transient backend error no longer forces the operator to mint a fresh code. `restore` drops codes already past their TTL. Added coverage for the restore path.

**P1 — Pairing limiter is optional** (`/api/pair`): left as-is by design. In every real deployment `IPLimiter` is non-nil: `cmd/frontdesk/main.go` always constructs `ratelimit.NewIPLimiter(...)` and passes it, and the nil-guard mirrors the sibling login ceremonies (webauthn/TOTP/OIDC) registered right above. The guard only elides the middleware in tests that omit a limiter. Independent of rate limiting, pairing codes are 80-bit random, single-use, capped at 20 outstanding, and expire in 3 minutes, so guessing is not the threat model. Happy to make the limiter mandatory at construction if you'd prefer fail-closed over the current parity with the other public auth routes.

PR #435, opened 2026-07-14
feat(bellhop): de-pill member detail, event range filters, infinite scroll, honest fleet sync

hugalafutro:
## What

Reworks the Bellhop member-detail screen and the surrounding event/sync plumbing.

- **De-pilled member detail**: ledger-style layout with flat `SectionHeader`
  dividers instead of stacked cards, sharing the event-row language with the
  Events screen.
- **Event range filters**: preset ranges (1h/24h/7d/30d/all) plus an absolute
  calendar range from a date picker, shared between the member-detail and
  Events screens (`EventRange` / `CustomDateRange` in `ui/common`).
- **Infinite scroll**: drift-proof windowed paging (re-request offset 0 with a
  bigger limit so an event arriving mid-scroll can't duplicate or skip rows),
  driven by a lazy sentinel that arms when the list bottoms out.
- **Honest fleet sync**: surfaces the real `last_sync_at` from
  `GET /api/fleet/autosync`, uses `context.WithoutCancel` for config sync, and
  a longer-lived sync client so a slow fleet sync isn't cut off.

## Review fixes (second commit)

- **Infinite-scroll retry backoff**: the sentinel re-arms the instant
  `loadingMore` clears, so a persistent fetch failure used to re-fire
  `loadMore` on every recomposition and hammer Front Desk. Both windows now
  count consecutive failures and back off exponentially (1s to 30s, shared
  `loadMoreBackoffMillis`), resetting on the first success. The delay runs
  outside the fetch mutex so the poll refresh is never stalled behind it.
- **Trailing divider**: the event-row divider now renders only between rows,
  not after the last one.
- **Comment fix**: the `CustomDateRange` doc claimed the server's upper bound
  was exclusive; `store_events` uses `created_at <= ?` (inclusive). Reworded to
  match, and it now explains why `untilRfc3339` still extends past midnight.
- **Test tidy**: `MemberDetailScreenTest` imports `EventRange` rather than
  fully-qualifying it.

## Testing

- `./gradlew testDebugUnitTest` green (full Android unit suite).
- `ktlintMainSourceSetCheck` / `ktlintTestSourceSetCheck` clean.
- New `LoadMoreBackoffTest` covers the backoff curve; new
  `loadMoreBacksOffAfterFailureInsteadOfHammering` proves the retry parks on
  its delay instead of re-hitting the client in the same tick.

### DIFF
--- a/android/app/src/main/kotlin/com/hugalafutro/bellhop/MainActivity.kt
+++ b/android/app/src/main/kotlin/com/hugalafutro/bellhop/MainActivity.kt
@@ -577,6 +577,7 @@ private fun LinkedContent(
             memberNames = ui.members.associate { it.id to it.name },
             onSeverity = eventsVm::setSeverity,
             onRange = eventsVm::setRange,
+            onCustomRange = eventsVm::setCustomRange,
             onLoadMore = eventsVm::loadMore,
         )
     } else if (showAlerts) {
@@ -638,6 +639,9 @@ private fun LinkedContent(
             onSyncFleet = { requireOperatorAuth { detailVm.syncFleet(ui.primaryId) } },
             onReconcile = { liveState -> detailVm.reconcile(liveState) },
             onDismissActionError = { detailVm.dismissActionError() },
+            onRange = detailVm::setRange,
+            onCustomRange = detailVm::setCustomRange,
+            onLoadMoreEvents = detailVm::loadMore,
         )
     } else {
         DashboardScreen(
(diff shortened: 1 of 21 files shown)


hugalafutro:
Split into two independent PRs so the backend can deploy before the app is verified against it:

- Backend (last_sync_at + disconnect-proof config sync): #436
- Android (member-detail de-pill, range filters, infinite scroll): #437

Closing this combined PR in favour of those.

PR #586, opened 2026-07-30
feat(discrepancies): provider accordion, bulk dismiss, and stop calling a dismissal a comeback

hugalafutro:
Reworks the claims half of the model discrepancy modal (the one behind the Models sidebar badge) from a flat wall of model pills into a navigable accordion, and fixes a reporting bug that made every dismissal read as "the model came back".

Frontend only. **No backend, migration or API change** — verified rather than assumed, see below.

## Why

A provider with 42 gone models rendered 42 mono pills, three buckets deep, for every provider at once. On the live fleet that is 179 rows on sight across 8 providers. Clearing a real backlog took ~70 individual clicks, because the only bulk control was a global Retest all.

Separately, dismissing a model announced it as **"is listed again"**, which is the opposite of what happened.

## The reporting bug

`dismissOptimistically` set `status: "resolved"` — the same value a model receives when the provider genuinely re-lists it. Once a provider's rows all cleared, the summary printed one line per model reading `resolvedPlain`: *"{{model}} is listed again"*. Dismissing 60 models produced 60 false lines.

A dismissed row is absent from the next refetch **because** it was dismissed (`listClaimRows` filters on `discovery_dismissed_at IS NULL`), so absence confirms the dismissal rather than contradicting it. A fourth `ClaimStatus` keeps the two apart. Stored state was always correct; only the reported reason was wrong.

## Navigation

Providers render as collapsed pills carrying their **actionable** counts. Click a pill for its bucket lines, click a line for the models. Only one provider and one line are ever open — one `openPath` atom holds both halves, so switching provider collapses the previous one and its line in a single move rather than leaving three handlers to uphold the invariant.

## Performance

Unrolling a 52-row bucket stuttered audibly on a high-end machine. Two causes:

- The animated `grid-template-rows: 0fr → 1fr` forces a full-subtree relayout on **every frame** of the transition.
- Collapsed rows were mounted only to give that transition something to animate.

Both gone. Rows mount only while their bucket is open, which the single-open rule makes free.

| | before | after |
|---|---|---|
| rows mounted, all collapsed | 179 | **0** |
| modal DOM nodes, all collapsed | ~1100 | **323** |

Measured on the built container, click to second animation frame: open a provider 32ms, open a 32-row bucket 31ms, switch to a 52-row bucket 25ms, close 26ms.

Rows are also de-pilled (no per-row border/background/rounding, one hairline divider owned by the container, meta on the same line as the id), following Bellhop's event list. Unmounting additionally beats the old `inert` trick for screen readers: a row that does not exist cannot be announced under an `aria-expanded="false"` toggle. The region wrapper still renders so `aria-controls` always resolves.

## Controls

| Control | Scope |
|---|---|
| Retest | one provider |
| Dismiss all | one provider: every actionable gone + stale id, one request, confirmed |
| Dismiss all (header) | every listed provider: one confirm, one refresh, one toast |
| Clean (broom) | appears only once nothing is actionable; drops the pill from view, writes nothing |

Suspect rows are never sent: `setModelsDismissed` only touches `enabled = false` rows, so a still-enabled model would be refused and undercount `updated`. The control is disabled with a tooltip saying why rather than hidden.

Verified end to end against the fleet: badge `75` → confirm → 8 requests, 179 ids, `claim_count` 75 → 0.

## Copy

"Stop tracking" read as "stop discovering". Dismissing does not touch discovery at all — it clears rows from a list, and `models.Upsert` clears `discovery_dismissed_at` on any sighting. Both dismiss tooltips now lead with "Clear … from this list" and say discovery keeps running. Chips gained tooltips, because "Gone" does not say gone *where*.

## Why no backend change

Three things verified before scoping this as frontend-only:

1. `POST /api/discovery/dismiss` already takes `model_ids` as an array and applies them in one `UPDATE … model_id = ANY($2)`.
2. Stale rows satisfy that statement's `enabled = false` predicate, so gone + stale needs nothing new. Suspect does not.
3. Both journal queries end `ORDER BY detected_at DESC`, so the new collapsed Recent-changes summary needs no new field.

## Bugs found and fixed during development, not shipped

- **Return-to-top never appeared.** The `IntersectionObserver` watched the whole provider `<section>`, and an unrolled section keeps intersecting long after its header leaves view. Unit tests could not catch it (a stubbed observer reports whatever you tell it), so the test now captures the `observe()` target.
- **A cleaned provider stayed hidden** even when a refresh gave it a genuine new claim — the false-reassurance class this rework exists to remove. The filter is now self-invalidating: hidden only *while* `providerHasNoPending` holds.
- **The modal-wide undo failed silently.** Now reports like the per-provider path. Verified by reverting the fix and confirming the test fails.
- **The badge said 75 while the confirm asked about 179** (badge counts gone; Dismiss all also clears stale). The confirm now names what its number covers.
- **A plural bug in my own English copy:** "across {{providers}} providers" pluralises independently of `{{count}}`, rendering "across 1 providers".

## i18n

30 keys across 29 locales, hand-translated. Counted keys carry the categories `Intl.PluralRules` reports per locale — `few`/`many` for cs/pl/ru/sk/uk/ro, `zero`/`two`/`few`/`many` for ar, `two` for he. A missing category renders **English silently** rather than failing, and `make i18n-check` compares locales to `en.json` rather than checking `t()` calls, so neither gap would have been caught.

## Docs

New `## The Model Discrepancy Modal` section in `wiki/Model-Discovery.md`, plus a `discrepancy_modal` task in the (gitignored) screenshot toolkit. While there, corrected a stale paragraph that still claimed opening the badge acks the recorded changes — #583 removed that.

## Also in this PR

Commit 1 is unrelated README work (Front Desk quota strip in the HA thumbnail, Bellhop gallery trimmed to three shots, Star History block removed since the service returns no image after GitHub's API changes). It predates the modal work and is kept here deliberately rather than split into a second PR.

## Verification

- 5588 frontend tests pass (291 files), `pnpm exec tsc -b` clean, `pnpm lint` clean, `make i18n-check` green across all three targets.
- Every behaviour above hand-verified on a real 8-provider / 75-claim instance, not only in jsdom.


hugalafutro:
## Response to review

**Fixed: unconfirmed partial dismissal (P1, `Layout.tsx`).** Valid, and a hole in my own reasoning. `useDiscrepancies`' `refresh()` absorbs its error and returns `undefined` rather than throwing, so "the refresh underneath is the reconciliation" only held when the refresh *succeeded*. A short `updated` cannot say which ids the server skipped, so a failed refresh left every optimistic row struck through — including models still actionable on the server, hidden from the counts and controls. That is the modal failing at its only job. An unconfirmed partial now rolls the optimistic write back and offers no Undo, since there is no optimistic write left to undo. Regression test verified by reverting the fix first.

**Accepted, not fixed: stale Undo can clear a newer dismissal (P1 ×2).** Both reports are technically correct — `dismissed: false` is an unconditional clear, so an Undo is last-write-wins.

I did implement a per-model generation compare-and-swap for this, and then removed it. Two reasons:

1. **Disproportionate.** This modal says "this model is gone; Retest or Dismiss". The worst case of an Undo clearing the wrong dismissal is that the modal nags about a model already dealt with, and one Dismiss click clears it again. No data loss, no routing effect. The fix was ~100 lines of ownership bookkeeping across three handlers plus a new user-facing string in 29 locales.
2. **It introduced the second finding.** The completion-order bug you found was in my fix, not in the original code. Answering complexity with more complexity is the signal to stop.

So this is a recorded trade-off rather than an oversight. If it ever proves worth closing properly, the right fix is server-side and conditional (clear only `WHERE discovery_dismissed_at` matches what we wrote), not client-side ownership tracking — a client-side scheme cannot see a dismissal from another tab or operator anyway, so it was never going to be airtight.

**On reachability, for the record.** The first stale-Undo repro does not work through the UI: two dismissals of the *same* model cannot leave two Undos on screen, because the toast provider de-duplicates identical messages and the second "Dismissed a" replaces the first. The reachable path is a batch toast plus a per-row toast (different messages, both clickable). T-Rex driving the API directly proves the API is last-write-wins, which is true, but not that the UI can get there. Useful distinction for weighting severity.

**Coverage.** The gate failure was mostly not this feature's code (97–100%) but `Layout`'s per-provider Dismiss all handler from an earlier commit, which had **no integration test at all** — the component tests only assert its callback fires, leaving its request, toasts, rollback and undo unexercised. Eight cases added. Diff coverage 83.9% → **95.1%**.

hugalafutro:
## Correction to my previous comment

That comment said the partial-dismiss reconciliation was fixed. It isn't, and I've removed that fix too. Superseding it here rather than leaving a claim I've contradicted.

**Where all three findings actually land: none are fixed by code, and I think that's right.**

### Partial dismissal + failed refresh — resolved as already-reported, not fixed

My rollback was wrong twice over.

It **did not reliably work**, which is the third P1: `revertDismissal` compares on `optimisticFrom`, and *any* merge strips that marker because a refetch is newer authority — deliberate since #583. A successful refresh landing mid-flight silently turned the rollback into a no-op, so it corrected the failure mode only when nothing else was happening. Good catch.

It was also **never needed**. `refresh()` records its failure, the hook exposes `refreshError`, `Layout` feeds it to the modal's `loadError`, and the modal renders a `role="alert"` banner over the list. A failed refresh was never silent — the modal already had a vocabulary for "we could not find out", which is precisely what #583 built that banner for. I should have checked what existed before adding a mechanism. The shortfall toast names the count, the banner says the state is unconfirmed, and the test now asserts the banner rather than a row rewrite that only sometimes happened.

### Stale Undo clears a newer dismissal — accepted trade-off

Technically correct: `dismissed: false` is an unconditional clear. I implemented a per-model generation compare-and-swap and removed it, because it was disproportionate and because it introduced the completion-order bug you then found. Worst case here is that the modal nags about a model already dealt with, and one Dismiss click clears it. No data loss, no routing effect.

This is a nag-suppression feature: it says "this model is gone, Retest or Dismiss". ~100 lines of ownership bookkeeping across three handlers, plus a user-facing string in 29 locales, is not the right price for that. If it ever is worth closing, the fix belongs server-side and conditional (`WHERE discovery_dismissed_at` matches what we wrote) — a client-side scheme can't see another tab or operator anyway, so it was never going to be airtight.

### Completion-order ownership — dissolved

Was a bug in my fix, not in the original code. Gone with the machinery.

---

**Net over both rounds: ~230 lines removed, ~20 added.** The handlers are close to what they were, plus the confirm dialog, the shortfall toast, and eight tests for `Layout`'s per-provider Dismiss all handler, which had **no integration test at all** — the component tests only assert its callback fires. That was the genuinely useful thing to come out of this review. Diff coverage 83.9% → **98.7%**, `Layout.tsx` at 100% of changed lines.

One factual note for severity weighting: the first stale-Undo repro does not work through the UI. Two dismissals of the *same* model cannot leave two Undos on screen, because the toast provider de-duplicates identical messages. The reachable path is a batch toast plus a per-row toast. Driving the API directly proves it is last-write-wins — true — but not that the UI can get there.

hugalafutro:
## Fixed in `5ad5bf5` — you were right and my banner argument was wrong

The detail that settles it is one I had underweighted: **the provider switches to `Clean`**. With every requested row marked dismissed, `providerHasNoPending` goes true, so the pill swaps Retest and Dismiss all for the broom — the operator loses the controls for precisely the models the server never dismissed. A refresh-error banner does not give those controls back. That is a materially worse failure than "a stale row is displayed", and my previous comment leaned on the banner as if it covered it. It didn't.

**The fix is to stop claiming what the server didn't confirm**, rather than to claim it and correct afterwards.

The bulk paths are no longer optimistic. A full-count response confirms every id, so the rows are marked; a short one confirms none of them *in particular*, so the rows stay actionable and the refresh reconciles. The modal-wide path applies the same rule per batch, where a rejected batch and a short batch are both just "not confirmed".

This is **smaller than what it replaces** — 59 lines out, 51 in, and `restoreClaim` drops out of both dependency lists. Marking up front and correcting afterwards could not be made to work, for the reason your earlier finding identified: only a successful refresh knows which ids took, `refresh()` absorbs its own failure, and the rollback compared on `optimisticFrom`, which any merge strips. Not making the claim removes the failure mode instead of reporting it.

Cost: no strike-through for one round trip, behind a confirmation dialog. Fair trade for never showing a provider as clean when it isn't.

The per-row path keeps its optimistic write — one model per request means `updated: 0` throws, so there is no partial state to misread, and instant feedback matters most there.

**Test asserts the consequence, not the banner:** after a short response with a failed refresh, no `Clean` control, Retest and Dismiss all both still present, rows still `pending`. Verified by reinstating the over-claim, which makes the broom appear and the test fail.

Diff coverage 98.7%, `Layout.tsx` at 100% of changed lines, 5595 tests passing.

### Still an accepted trade-off: stale Undo

Unchanged and deliberate. An Undo is last-write-wins and a superseded one can clear a newer dismissal; recovery is a single Dismiss click. I built the per-model generation compare-and-swap for this and removed it, because it introduced the completion-order bug you then found, and because ~100 lines of ownership bookkeeping plus a string in 29 locales is the wrong price for a nag-suppressor. If it ever is worth closing, the fix belongs server-side and conditional (`WHERE discovery_dismissed_at` matches what we wrote) — a client-side scheme can't see another tab or operator, so it was never going to be airtight.

PR #592, opened 2026-07-30
fix(provider): audit hardcoded model catalogs for sunsetted and missing models

hugalafutro:
Audits all ten embedded model catalogs in `internal/provider/catalogs/` against three
independent sources: models.dev, each vendor's own deprecation page, and live requests
issued through a running gateway. Two commits: removals first, then additions.

## Why the rot was invisible

Five catalogs (`xai`, `zai`, `deepseek`, `opencode_go`, `opencode_zen`) union into discovery
via `mergeLiveAndCatalog`, so every catalog entry is re-marked as "seen" on every scan.
Discovery's stale/gone detection therefore can **never** flag a catalog entry whose upstream
model has died. That is the systemic reason the entries below sat rotting.

## Commit 1: removals (19 entries)

Only entries confirmed dead by at least two of the three sources.

**`xai.json` (8) is the one with real user impact.** xAI retired these on 2026-05-15 but
kept the slugs resolving, so probes return HTTP 200 and nothing looked wrong. The response
`model` field gives it away: requests silently redirect to `grok-4.3` (`grok-build-0.1` for
`grok-code-fast-1`) and bill at grok-4.3's unified $1.25/$2.50.

| Catalog entry | Actually served | Catalog price | Real price | Error |
|---|---|---|---|---|
| `grok-4-fast-reasoning` | grok-4.3 | $0.20 / $0.50 | $1.25 / $2.50 | 6.25x under |
| `grok-4-fast-non-reasoning` | grok-4.3 | $0.20 / $0.50 | $1.25 / $2.50 | 6.25x under |
| `grok-4-1-fast-reasoning` | grok-4.3 | $0.20 / $0.50 | $1.25 / $2.50 | 6.25x under |
| `grok-4-1-fast-non-reasoning` | grok-4.3 | $0.20 / $0.50 | $1.25 / $2.50 | 6.25x under |
| `grok-code-fast-1` | grok-build-0.1 | $0.20 / $1.50 | $1.25 / $2.50 | 6.25x under |
| `grok-3-mini` | grok-4.3 | $0.30 / $0.50 | $1.25 / $2.50 | 4.2x under |
| `grok-3` | grok-4.3 | $3 / $15 | $1.25 / $2.50 | 2.4x over |
| `grok-4-0709` | grok-4.3 | $3 / $15 | $1.25 / $2.50 | 2.4x over |

All eight also claimed a 131072 context against grok-4.3's 1000000.

**`google.json` (2).** `gemini-2.0-flash` and `gemini-2.0-flash-lite` passed their
2026-06-01 shutdown and return `404 "This model is no longer available"`, but Google still
lists them in `/models`, so they were discovered as healthy and offered to callers.

**`opencode_zen.json` (6).** `gemini-3-pro`, `glm-4.7`, `grok-code`, `kimi-k2-thinking` and
`qwen3-coder` return `"Model X is not supported"`; `claude-sonnet-4` returns `"not supported
on the full model list"`. Because this catalog unions into discovery, all six were
advertised as routable.

**`anthropic.json` (2).** `claude-opus-4` and `claude-sonnet-4`, retired 2026-06-15. This
catalog is pricing-only, so the rows were already inert. Dead-weight cleanup only.

**`zai.json` (1).** `glm-4-32b-0414-128k`, which Z.ai's docs no longer list anywhere.

**`vertex_express.json` (2).** The two shut-down 2.0 models.

## Commit 2: additions (28 entries)

Highlights: `claude-opus-5` and `claude-sonnet-5` were missing from our own Anthropic
pricing table, `grok-4.5` from xAI, the whole `gpt-5.6` family from OpenAI, and nine live
Google models had no pricing at all. Full per-file breakdown is in the commit message.

## Deliberately not changed

Three findings that look like sunsets but are not, recorded so they do not get
re-litigated:

- **Z.ai `glm-4.5-x`, `glm-4.5-airx`, `glm-4.6v-flash`, `glm-4.6v-flashx`** return
  `"Insufficient balance or no resource package"`, not a not-found, and Z.ai's docs still
  list them. They are pay-as-you-go models outside the coding plan. Removing them would
  break anyone holding a balance.
- **OpenCode Zen `glm-5/5.1/5.2`, `deepseek-v4-pro/flash`, `minimax-m2.5/m2.7`** return
  `"Upstream request failed"`, a transient aggregator fault. glm-5.2 is their flagship.
- **All `gpt-5.x` on OpenCode Zen** initially looked dead; that was my own probe artifact
  from `max_tokens: 1` producing an empty completion. They pass at 64 tokens.
- **`gemini-2.5-flash-native-audio-preview-12-2025` and `gemini-3.1-flash-live-preview`**
  are active per Google with no shutdown announced. They are missing from discovery because
  Live API models are not in the listing we scan, not because they died.
- **`gemini-embedding-001`** is past its earliest shutdown date (2026-07-14) but is
  verified **alive**: probed through `/v1/embeddings` it returns a 3072-dim vector. Google
  documents its shutdown dates as *earliest possible* dates, and this one has not been acted
  on. Kept, and it should not be removed on the strength of the docs alone.
- **Cohere's `tiny-aya-*`** are live but unpriced by Cohere. Discovery applies a catalog
  entry's prices unconditionally, so adding them would assert they are free.

## Two follow-ups this audit surfaced (not fixed here)

1. **Upstream error messages are dropped from client responses.** Everything collapses to
   `"upstream provider returned HTTP 400"`. The real message survives only in the request
   log, which is the sole reason I could separate "retired" from "out of credits" from
   "transient fault". Those need different operator responses.
2. **OpenCode Zen's Gemini models are unusable through the gateway.** `gemini-3-flash`,
   `gemini-3.1-pro` and `gemini-3.5-flash` fail with a Gemini-native `Invalid JSON request
   body: Missing key at ["contents"]`: we post an OpenAI-shaped body to a native-Gemini
   route.

## Testing

`go test ./internal/provider/...` passes (904 tests) and `golangci-lint` is clean.
`TestAnthropicPricingLookupDated` was updated: the two retired dated Claude IDs now assert
*not found* instead of asserting pricing. Date-stripping lookup stays covered by the four
remaining dated IDs. No production Go lines changed, so the diff-coverage gate is a no-op.


hugalafutro:
Both P1 findings addressed in b4219d21. The diagnosis was correct and matches what I saw
probing production: `google.json` is pricing-only, so removing an entry skips price
enrichment and nothing else. Google still returns both 2.0 models from `/models` with
`generateContent`, so discovery kept building them with `Enabled: true`.

Fix: `catalogs/google_retired.json` + `IsRetiredGoogleModel`, applied in the discovery loop
beside the existing non-chat filter. It also covers `gemini-2.0-flash-001` and
`gemini-2.0-flash-lite-001`, which Google still lists and which were never in the pricing
catalog, so the earlier commit could not have touched them.

Regression test added as requested: `TestDiscoverGoogleAIStudio_FiltersRetiredModels` feeds a
listing containing two retired IDs plus one current model and asserts only the current one
survives. Diff coverage on the new code is 100% (6/6).

One deliberate divergence from the suggested fix. I did not gate discovery on a
pricing-catalog match or an allowlist. A model absent from `google.json` is normally just
unpriced, not unusable, and Google ships new models faster than we catalogue them, so
requiring a catalog hit would silently drop working models. The denylist is also explicit IDs
rather than a `gemini-2.0-` prefix match, because Google publishes shutdown dates as the
*earliest possible* date and does not always act on them: `gemini-embedding-001` is sixteen
days past its date and still returns 3072-dim vectors. An ID earns a place on this list only
once a real request has been observed to 404.

PR #593, opened 2026-07-30
fix(proxy): classify upstream errors, fix OpenCode Zen routing, auto-disable retired models

hugalafutro:
The two follow-ups from the catalog audit (#592), plus the work they turned up. Every claim
below was verified against a live provider or on the dev stack; nothing here rests on docs
alone.

## What was broken

**Upstream errors were indistinguishable.** Every failure recorded `provider_error` and told
the caller `upstream provider returned HTTP 400`. Three failures needing completely different
responses looked identical: a model the provider retired, a model the account cannot pay for,
and a transient fault. Working out which was which required reading
`request_logs.error_message` by hand — that is how #592 was done.

**OpenCode Zen's Gemini models were unusable.** Zen routes each family to a different upstream
dialect and has no OpenAI-compatible route covering all of them, so an OpenAI-shaped body
returned Google's own `Missing key at ["contents"]`.

**We were breaking most of Zen ourselves.** MH injected `chat_template_args` into every Zen
request. A/B testing each model showed `glm-5`, `glm-5.1`, `glm-5.2`, `minimax-m3`,
`kimi-k2.6`, `kimi-k3` and `deepseek-v4-pro` all fail with it and succeed without it. Only
`qwen3.6-plus` tolerated it. Zen's error is too generic for the param-strip retry to
recognise, so those models were simply dead through the gateway.

**Keyless Zen discovered nothing.** Keyless discovery emits only live models with a
zero-priced *catalog* entry, and not one of Zen's six free models was catalogued.

**gpt-5.x context was ~4x too small.** The catalog stored `272000` as the context window. Per
OpenAI's docs that is a 1,050,000 window with 128,000 max output; 272K is only the threshold
above which billing switches to 2x input / 1.5x output.

## What changed

**Error classification.** `classifyUpstreamError` maps status + sanitized body to a stable
kind, adding `provider_model_gone`, `provider_not_entitled` and `provider_bad_request`. Every
pattern is cited to a real provider response. Strictly observability: failover, breaker trips
and quota handling stay status-code driven, because the existing design deliberately funnels
billing failures into the 429 path so failover moves on.

The provider's own body is still never forwarded — it can quote the request back at us — and
a test plants a secret in an upstream error to prove neither the reason nor the client
message can leak it.

**Auto-disable on `provider_model_gone`.** Three consecutive refusals with no success in
between disable the model and emit `model.auto_disabled_gone` (alert catalog, Discovery,
default-on, translated into all 29 dashboard locales). This closes a gap discovery cannot
reach: `RecordMissingModels` only fires when a model *leaves* a listing, and Google served
`gemini-2.0-flash` from `/models` for two months after shutting it down. Strikes are
in-memory and per-instance, so the hot path takes no database write and nothing fans a
disable across the fleet on one member's evidence.

**Zen Gemini routing.** Reuses the existing vertex-express egress adapter rather than adding
a second translation layer; only the path (`/models/{id}:generateContent`, no Vertex
publisher prefix) and the auth (`x-goog-api-key`; Zen answers Bearer with `AuthError`) became
provider-aware.

**Catalogs.** Zen and Go now match what each can actually serve (59/59 and 20/20). The
pricing catalogs were shrunk from mirrors to override channels — a row that merely restated
models.dev is deleted, since the catalog *wins* over models.dev and a stale duplicate
silently overrides correct data. That is exactly how xAI's retired models metered 6x wrong.
187 rows → 128.

## Deliberately not done

- `claude-sonnet-4` is in Zen's listing and on its website but a real request returns "not
  supported on the full model list". Not re-added.
- `hy3-preview`, `mimo-v2-omni`, `mimo-v2-pro` are listed by Go and refuse. Dropped, not added.
- Z.ai's `glm-4.5-x`/`airx`/`4.6v-flash`/`flashx` return "insufficient balance", not
  not-found, and Z.ai's docs still list them. They are pay-as-you-go models outside the
  coding plan, not retired. Left in place.
- `gemini-embedding-001` is past its published shutdown date but verified alive via
  `/v1/embeddings` (3072-dim vector). Google's dates are earliest-possible and it did not act
  on this one. Kept.

## Review response

All three findings fixed rather than documented, in `7489dc1`: the bare `billing` substring
(a transient `billing_engine_timeout` on a 500 was recorded as a permanent entitlement
failure — 402 now classifies on its own instead); the streaming asymmetry (worse than a
missing call — `noteModelServed` fired on the 200 headers, so a 200-then-gone stream cleared
its own streak and could never accumulate one; the verdict now defers to the end of the
stream); and `does not exist` now requiring the body to mention a model, so an error about
another entity cannot accrue strikes against a live model.

## Verification

6026 Go tests, `golangci-lint` clean, `make i18n-check` green across web/fd/android, `tsc -b`
clean, 1656 frontend tests.

End-to-end on the dev stack, after a rebuild:

- Zen `gemini-3.6-flash`, `gemini-3.5-flash`, `gemini-3.1-pro`, `glm-5.2`, `kimi-k3`,
  `minimax-m3`, `deepseek-v4-pro` all answer. Every one of those was failing before.
- Streaming through the Gemini egress produces correctly translated chunks.
- Go still returns reasoning content, confirming its injection is intact.
- Discovery's retired filter drops the `gemini-2.0-*` rows: live models got the new scan's
  timestamp, those kept the old one.
- Three refusals of a genuinely dead model flipped it to `enabled=false`, logged
  `error_kind=provider_model_gone`, and emitted the event on the SSE stream with
  `strikes: 3`. Neighbouring models stayed enabled, confirming strikes are per-model.


hugalafutro:
All three addressed in 5bb79d6. The second one was a defect in my own previous review fix, so
thank you for catching it.

**1. Capability errors could auto-disable a live model.** Correct, and the inconsistency was
mine: the previous commit fixed exactly this shape for `does not exist` by requiring model
context, then left the neighbouring `is not supported` bare.

Requiring the word "model" anywhere is not sufficient either, because the parameter-error
shape contains it (`top_p is not supported for this model`). The generic phrases now match
only when a model is named *before* them on the same line, which is how every real payload
reads and which the trailing-model shape fails. Phrases that name a model inherently
(`model not found`, `unknown model`, `not supported on the full model list`) still match
anywhere.

Negative tests added for all four categories you named — parameter, operation, region,
account capability — plus non-model entities (`conversation does not exist`, `uploaded file
does not exist`, `API version is no longer available`). Six real retired-model payloads assert
the tightening did not break the positive direction.

**2. Failed streams cleared the strike streak.** Right, and worse than the symptom suggests. My
fix used a two-way split, so every non-gone outcome counted as "the model answered", including
transient errors, disconnects and stalls. A retired model would reset its own streak with its
own failures and stay routable forever — the precise failure the feature exists to prevent.

The rule is three-way and now says so explicitly: only a clean finish clears, only an explicit
gone-report strikes, anything else is inconclusive and touches neither counter. Lifted into
`verdictForStream` so the policy is named and tested rather than an inline condition that has
now been got wrong twice.

**3. Undersized custom failover groups.** Valid. Discovery already revalidates after disabling
a model; this path did not, so a group could sit enabled with fewer than two routable members
until an unrelated scan fixed it. It now calls the same repository method, best-effort so a
failure cannot undo the disable.

This one has no unit test, which I want to be explicit about rather than let the test count
imply coverage: it needs a live database, which is the same reason T-Rex could not reach its
assertions. It is verified only by using the identical call discovery uses.

Re-verified end to end on a rebuilt dev stack: Zen's Gemini models still serve, and three
refusals of a genuinely dead model still disable it.

hugalafutro:
Both fixed in b654f3a. Same class as the first round — the served/gone policy was not applied
consistently across every streaming path — so this time it is a single shared helper rather
than a second copy of the logic.

**1. Empty streams cleared the streak.** Correct. An absent error kind was being read as proof
the model answered, but a stream can open, emit nothing and end without recording an error;
with the TTFT probe disabled nothing else stops it. Clearing now requires positive evidence
that content flowed. `streamProducedOutput` accepts either completion tokens or a non-zero
TTFT, because neither is reliable alone — usage is absent when a provider omits the usage
chunk, TTFT is zero when the probe is off. A stream that delivered nothing is inconclusive and
touches neither counter.

**2. Hedged winner recorded no verdict.** Also correct, and the more serious of the two: with
hedging enabled a model reported gone mid-stream never accrued a strike *or* cleared one. Both
dispatch sites now call one `noteStreamOutcome`, so the sequential and hedged paths cannot
drift again — which is exactly how this diverged.

Tests cover the full verdict table including truncation, both `producedOutput` signals
independently, and the composed cases in each direction: strikes surviving empty streams, and
a stream that really delivers content still clearing them.

Worth noting your validation was blocked both rounds — the proxy package's `TestMain` needs
PostgreSQL on 5433, which your environment does not have — so both findings came from reading
the control flow rather than executing it. Both were right, and the hedged path in particular
is one no test I had written would have caught.

hugalafutro:
Fixed in edd88b5. Valid, and this is the first round where your runtime repro actually
executed — good to see it disable a live model rather than reason about it.

The shape survived last round's tightening because the model *is* named before the phrase, so
the trailing-model guard (`top_p is not supported for this model`) never applied. What
distinguishes it is what comes **after**: a capability qualifier.

`modelCapabilityRefusal` now vetoes a match whose rejection phrase is followed by
for/with/on/in some operation, endpoint, method, route, request type, region, plan, tier or
account.

Two constraints shaped it, both worth recording:

- It has to be a separate veto rather than part of the pattern, because RE2 has no negative
  lookahead.
- It cannot be a blanket "any trailing text disqualifies" rule. Real retirement messages
  continue past the phrase too — OpenAI's is `...does not exist or you do not have access to
  it` — and Zen's `not supported on the full model list` is a retirement signal whose
  qualifier simply is not a capability.

Tests: nine operation/endpoint/region/plan shapes that must not classify, plus the three-strike
integration you asked for — repeated capability refusals never reaching `SetEnabled`, paired
with a genuinely retired model that still does.

That is three rounds on the same policy in three different places, so the cover is now
deliberately at both levels: the classifier table, and the composed classify-then-strike path
that a future loosening would have to get past. Re-verified on a rebuilt dev stack in both
directions.

hugalafutro:
Fixed in a049902. Correct, and it was a genuine race rather than a theoretical one — a retired
model is exactly the one drawing concurrent refusals, since clients retry it and a failover
group can reach it from several requests at once. Two racing strikes both read the same value
and stored the same increment, so the streak stalled below the threshold and the model stayed
routable.

The map now holds a per-model `*atomic.Int64`. `LoadOrStore` still races on *creating* the
counter, which is harmless since everyone increments the winner's instance, and the increment
itself is atomic. Only the caller observing exactly the threshold disables, so a burst issues
one disable rather than one per goroutine that saw the count at or above it.

Writing the burst test then caught a second defect, this one mine rather than reported:
clearing the counter at the threshold let the next refusal start a fresh count, so 50
concurrent retries would disable roughly every three strikes and publish an alert each time.
The count now stays above the threshold and later refusals are no-ops.

That inverts my original reason for clearing it — that a failed disable should be retried
rather than parked — so the failure path now clears it explicitly: a disable that could not be
written rebuilds its streak and tries again, a successful one never repeats. Covered by its own
test.

All three concurrency tests run under `-race`. Also verified against the real handler on dev:
re-enabled a genuinely dead model, fired eight concurrent requests, got exactly one
`auto-disabled retired model` line and one disable.

Noting for the record that your validation was blocked again this round (PostgreSQL on 5433) —
the concurrency finding still came out of reading the code and was right. Four rounds, four
real defects, two of which no test I had written would have caught.

hugalafutro:
Fixed in 074658d, and this one takes your structural suggestion rather than tuning the patterns
again — which was the right call, because four rounds of phrase-tightening were each
approximating the same missing constraint from a different angle.

`classifyUpstreamError` now takes the upstream model id for the attempt, and a retirement
verdict requires the provider's phrase to appear within 80 characters of **that id**. The
wording still participates but is no longer load-bearing on its own: the body has to be the
provider talking about this model.

Looking back at the sequence, every round was the same defect wearing a different hat — a
trailing model, then a trailing capability qualifier, then an unrelated entity, then an
unrelated model, now echoed request content. None of them could be fixed by better prose
matching, because the text was never required to be *about* the model in question. Anchoring on
identity covers all five at once. The capability-refusal veto stays on top, since
`Model X is not supported for this operation` names the right model and still is not a
retirement.

Two details worth recording:

- Both id spellings are handled. Providers report Google models as `models/gemini-2.0-flash`
  while the catalog stores the bare id, so the comparison is on the last path segment. Tested
  in both directions.
- An empty model id substantiates nothing, so it classifies nothing — the caller gets the
  generic provider error rather than a retirement that cannot be checked. That is the safe
  direction: a missed retirement costs a stale model, a false one disables a working model.

Tests for each shape you reproduced: an error naming a different model, a fallback chain
mentioning another dead model, a provider echoing prompt text containing `unknown model`, an
echoed prompt containing `model not found`, and the no-id case — against positive controls with
identical phrasing where the model IS ours.

6089 tests, lint clean. Verified on dev: Zen Gemini still serves, and three refusals of a
genuinely dead model still log `provider_model_gone` and disable it.

hugalafutro:
Fixed in 3aacb8e. Real race, and a consequence of detaching the disable to keep it off the
request path: the write can still be in flight, for as long as the database takes, when the
model answers and proves it is alive.

The strike counter becomes a `goneStreak` carrying the count plus a cancellation tombstone.
`noteModelServed` marks it cancelled; the queued disable checks that immediately before
writing and stands down.

The ordering is the load-bearing part, and is commented as such: mark cancelled **first**, then
drop the map entry. The queued goroutine holds a pointer to the streak rather than the map
entry, so marking before deleting is what makes the flag reach it — deleting first would leave
it holding an orphan it could never be told about.

Being straight about the limit: this does not close the window to zero. A success landing
between the check and the database commit still loses. It shrinks it from the whole round trip
to the gap between two adjacent statements. Closing it fully would mean either holding the
request path on the write, or re-enabling after the fact — and re-enabling risks fighting an
operator who disabled the model deliberately in the same window. I did not think either trade
was worth it, but say so explicitly rather than imply the race is gone.

Both tests run under `-race` with a gated mock, so the interleaving is deterministic rather
than timing-dependent: a success while the write is held open cancels it, and the same held-open
write with no success still lands.

Thanks for re-checking the capability-refusal finding against the current code and confirming
it is vetoed — useful to have that closed out rather than left ambiguous.

One note from verification, unrelated to this change: OpenCode Zen's `gemini-3.6-flash` is
currently returning its own `Internal server error` 500. The other four Zen Gemini models route
fine through the same egress path, so that is upstream rather than the routing.

hugalafutro:
Fixed in ceb315c. You reproduced the exact window I said I was leaving open, so it is worth being clear that I was wrong about the trade rather than pretending the previous commit covered it.

Both of your suggested remedies were on the table. Serialising is not available on this path: `noteModelServed` runs on the request path *before* a non-streaming response is written to the client (`proxy_failover.go:191`), so a lock held across the disable write would put one request's client latency behind a database write belonging to a different request's error path — and behind the full context budget if the database is slow, which is precisely when the lock would be contended.

So the write is made conditional at commit time in the only way left: the flag is read again *after* `SetEnabled`, and a disable that a success overtook is undone. The pre-write check stays, because skipping the write is better than reverting it whenever the timing allows; the second read exists for the case the first is by definition too early to see.

Three details worth recording:

- The undo runs on a **fresh context**. A slow write is exactly what opens this window, so `dctx` may have little or none of its budget left at the moment the compensation matters most.
- **No alert and no group revalidation** on that branch. The model ends up enabled, so publishing a retirement event for it would be false, and revalidating groups against a state that no longer exists is pointless work.
- The mock's gate now blocks only the *first* `SetEnabled`, otherwise a test holding the disable open would also deadlock the re-enable that follows it.

The new test waits on `setEnabledEntered` before calling `noteModelServed`, so the success is guaranteed to land inside the write rather than racing the goroutine's scheduling, and asserts the pair `[disabled, enabled]` in order. It runs under `-race` alongside the existing before-the-write case, which still asserts zero writes — the two halves of the window now have one test each.

What remains, stated rather than implied: an operator disabling the model in the microseconds between our two writes would have their disable reverted. That is strictly narrower than the bug being fixed and needs a success in the same instant to trigger at all, so I am accepting it rather than adding a read-modify-write.

6217 tests, lint clean.

hugalafutro:
Fixed in 0877f03. Valid, and it is my own reasoning applied inconsistently: I gave the compensating re-enable a fresh context in the previous commit for exactly this reason, then left `RevalidateCustomGroups` sharing the deadline the disable had already spent.

Each of the three out-of-band writes now takes its own `goneWriteTimeout` and cancels it immediately, rather than one budget being passed down the sequence.

The part worth stating explicitly is why a shared deadline is not merely untidy: the slowness is **correlated, not random**. The database load that makes a disable consume its whole budget is the same load the follow-up work has to get through, so the shared deadline expires precisely in the conditions where revalidating an undersized group matters most. That is what makes this a bug rather than a tuning question, and it is why the fix is per-write deadlines instead of a larger shared one. Nothing waits on the total — the work is already detached from the request path — so bounding each step costs nothing.

**Coverage, stated rather than implied.** The new test asserts the invariant on the compensating re-enable, which is the sequence the mock can observe. The revalidation call has the identical shape and the identical fix, but `failoverRepo` is a concrete `*failover.Repository` and needs PostgreSQL, so it is not covered by a unit test — same limitation I flagged when that call was added, and the same reason T-Rex could reach the reasoning but not the assertion. It holds a slow write open for 500ms and asserts the following write still has most of its budget; an inherited deadline shows up as `goneWriteTimeout` minus that delay and fails against a tolerance of half it.

**Two existing tests needed correcting, and one of them was wrong before this PR:**

- `SuccessCancelsAQueuedDisable` was asserting a branch the scheduler picks. Once a disable can be reverted as well as skipped, a success racing it can legitimately land before the pre-write check or after it. Both outcomes are correct and neither can be forced from outside, so it now asserts what must hold either way — the model is not left disabled — and the revert branch keeps its own deterministic test gated on `setEnabledEntered`.
- `FailedDisableIsRetried` began its second round as soon as the failing call was *recorded*, but the streak is cleared only after that call *returns*. In between, the second round could strike the old streak past the threshold, where every strike is a no-op, and then watch the first goroutine delete the evidence. Latent since I wrote it; this round's timing shift is what surfaced it. It now waits for the clear.

Ran the concurrency set at `-count=5` under `-race` to confirm the fixed tests are deterministic rather than merely passing once. 6218 tests, lint clean.

hugalafutro:
Fixed in 61ca11d. Correct, and I applied it to **both** cleanup paths rather than only the one you reproduced.

The failure path deleted from `goneStrikes` by model id, so a goroutine that had already been superseded could throw away a streak it never owned. Both deletes are now `CompareAndDelete` against the captured `*goneStreak`.

`noteModelServed` had the same window and I fixed it alongside: between its `Load` and its `Delete`, a failing disable can drop the streak it observed and new refusals can start another. That success is no evidence about a streak that began after it looked, so it must not clear it either. One reproduced instance, two call sites — worth closing both while the reasoning is in hand rather than waiting for the second to be reproduced separately.

**On the test, because the obvious version of it does not actually discriminate.** Counting disable attempts races the stale cleanup against the strike that completes the newer streak: if the strike wins, a second disable is attempted and the test passes *with or without* the fix. So it asserts the newer streak's **identity** instead — an unconditional delete removes the entry within microseconds of the write returning, which the observation window catches every time.

I confirmed that rather than assuming it: restoring `Delete(modelID)` fails the new test on `the stale failed disable deleted the newer streak`, and it passes again with `CompareAndDelete`. A regression test that has never been seen to fail is only a guess about what it covers.

6219 tests, lint clean, concurrency set green at `-count=3` under `-race`.

Worth noting what this round and the last two have in common, since it is the same defect class each time: every one has been a **detached goroutine acting on state that moved under it** — a cancelled disable, a spent deadline, now a replaced streak. The pointer captured at spawn time is the only thing that is still valid when it wakes up, so the code now treats the map entry as something to verify rather than something it owns.

hugalafutro:
Both fixed in 608494f.

**1. Overlapping model ids.** Correct, and the reason it matters is structural rather than incidental: model families are named by *extension*, so the collisions are systematic — `gpt-4` inside `gpt-4.1`, `gemini-3-flash` inside `gemini-3-flash-lite`. Errors about a newer sibling are exactly the errors a gateway sees while a family is being rolled forward.

The id now has to match as a whole identifier. Two membership choices worth recording, because they are the whole substance of the fix:

- `/` is **not** an id character. Providers path-qualify the same model (`models/gemini-2.0-flash`, `openai/gpt-4`), so a slash neighbour still refers to our model and must not disqualify it.
- `:` and `@` **are**, because they pin a distinct variant (`llama3:8b`, `text-bison@001`). Treating them as separators would recreate the bug for Ollama tags and Vertex versions.

Nine cases, including the reverse direction you did not raise — requested `gpt-4.1` against an error naming `gpt-4` — plus positive controls for backtick, path and vendor-prefixed spellings so the tightening cannot silently kill the true positives.

Writing them caught one of my own: two `llama3` cases used a body with no retirement verb in it, so they passed without exercising the boundary at all. A test that passes for the wrong reason is worse than no test.

**2. Rollback leaves the group disabled.** Also correct, and your suggested remedy would not have worked — worth flagging rather than quietly doing something else. `revalidateCustomGroups` only ever *disables*: it skips groups already disabled and `continue`s on `count >= 2`. Calling it after a rollback is a no-op, and nothing anywhere re-enables an auto-disabled group, deliberately — that is an operator decision surfaced through the discovery-claim badge.

So the fix is to stop producing the intermediate state instead of correcting it. `SetEnabledIfConfirmed` stages the write in a transaction and commits only if the justification still holds; an abandoned disable is never visible to any other session, so no revalidation can derive a group disable from it.

This inverts my previous answer to your round-7 finding, which is the honest summary: compensating after the fact was the wrong shape, because the damage was never confined to the row I was compensating.

The post-commit rollback stays for the one window staging cannot close — a success landing after `confirm` returns while the commit is in flight — since a committed write can only be undone. That path still has the brief visibility, and I would rather keep it than leave a model disabled after it proved it works.

**The database test is the load-bearing one.** It reads the row from a separate session *while the write is staged* and asserts it still sees the old value. No mock can demonstrate cross-session visibility, so the mock-based tests here would have passed just as happily against the broken version.

6231 tests, lint clean, concurrency and classifier sets green at `-count=3` under `-race`.

hugalafutro:
Fixed in 4025a96. Correct, and it is the mirror image of the bug that introduced this veto — worth naming, because the two together explain what the rule should have been all along.

Round 3 added the veto because a capability refusal was being read as a retirement. It was applied to the whole body, so it then did the opposite: any capability refusal anywhere cancelled a retirement stated elsewhere in the same response. Both are the same mistake — judging one phrase by text that belongs to a different one.

The pattern is now anchored and evaluated at the position of the phrase that matched, next to the existing per-phrase model-id check. Both halves of the verdict are local to one phrase, so an unrelated sentence can neither create it nor suppress it, and a mixed response is judged on whichever phrase actually describes the requested model.

Five cases, deliberately including the one your report did not cover: our model refused a *capability* while a **different** model is the one retired. That must stay a generic provider error, and a naive "just drop the veto" fix would have turned it into a retirement of a healthy model — which is how this whole sequence started.

Also covered: the refusal appearing before the retirement rather than after, and both phrases naming our model, where the retirement wins because the provider did assert it.

I confirmed the tests discriminate rather than assuming it — restoring the whole-body veto fails three of the five, and the nine operation-refusal shapes from round 3 all still pass, so the tightening did not simply undo it.

6245 tests, lint clean, diff coverage 93.0% locally.

hugalafutro:
Both fixed in fb28ea9.

**1. Nearby clause is not attribution.** Correct, and it is the honest reading of what the window check ever was: proximity standing in for grammar. `namesModelID` asked "does our id appear near this phrase", which a response naming our model in one clause and retiring a different one in the next satisfies perfectly.

The id must now be *bound* to the phrase: no clause boundary and no competing model id between them, within a short gap. Both sides of the phrase are allowed, because provider wording splits along grammatical lines — predicates take the id before them (`gpt-4 does not exist`), noun-phrase verbs take it after (`unknown model openai/gpt-4`). Restricting to one side would have broken half the real payloads.

Two judgement calls worth recording:

- A **comma** counts as a clause break. `healthy-model was selected, but retired-model does not exist` is two claims. This costs the phrasing `The model gpt-4, which you requested, does not exist` — a false negative, which is the safe direction: a missed retirement leaves a stale model, a false one disables a live one.
- "Competing id" is a token containing a digit or a hyphen. That catches `retired-model`, `gpt-4.1` and `llama3` while letting vendor prefixes like `openai/` through, which is what keeps the noun-phrase form working.

All 70 pre-existing classifier cases still pass, so this is a tightening rather than a rewrite. Eight new cases; the five negatives all fail against the old window check.

**2. Automatic disable becomes permanent.** Also correct, and worse than the title suggests — it is permanent *and invisible*. `Upsert` re-enables on re-sighting only while `disabled_manually` is false, and `listGroupClaims` hides models where it is true. So the provider could restore the model and it would stay off, with nothing surfaced for an operator to act on.

The fix is to match what the codebase already does: `RecordMissingModels` disables a vanished model with `enabled = false` and deliberately leaves `disabled_manually` alone. My write had diverged from that for no reason other than reusing SetEnableds SQL. The flag records an operators decision and is not the gateways to claim; the auto-disable alert is what announces this.

The test goes through a real `Upsert` rather than asserting the flag directly, because the flag only matters for what it makes the next sighting do — asserting the column would have pinned the mechanism instead of the behaviour.

Noting your T-Rex run could not reach PostgreSQL on 5433 again, so this one came from reading the schema usage rather than executing it. It was right, and it is the finding in this PR I would have been least likely to catch myself: nothing in the proxy package hints that a boolean on the model row governs whether discovery can ever bring the model back.

6255 tests, lint clean, diff coverage 93.1%.

hugalafutro:
Both fixed in 973eb73, and the two of them together are one bug, which is the useful thing this round established.

**`enabled` plus `disabled_manually` can express two kinds of disable, and there are three.** That is why the last two rounds traded one failure for another: stamping the flag made an automatic retirement permanent and invisible; leaving it clear let the next `Upsert` revive the model. Both of your findings are correct and neither could be fixed without the third state, so I stopped patching the flag and added one.

Migration 063 adds `models.auto_retired_at` — the same distinction migration 062 already draws for failover groups, for the same stated reason: the flag alone cannot answer *who did this*.

The three now read distinctly:

- **Operator disable** — never undone automatically, never nags.
- **Discovery disable** — SHOULD be revived by a re-sighting; the model had vanished from the listing, so its return is genuine new evidence.
- **Traffic retirement** — revived by neither. The model never left the listing; the provider was refusing it while still advertising it, so a sighting proves nothing. Reviving on one puts it back into routing to fail, re-alert and churn groups every scan, exactly as you describe.

Every operator-driven write of `enabled` clears the stamp, so enabling by hand is how an operator tells the gateway to trust the listing again.

**On the second finding**, you are right and I was wrong to keep waving it through — I called it an accepted residual twice on the grounds that the window was vanishingly small. Size was never the point: the loss is an operators deliberate decision, silently. The undo is now a conditional update that only fires while the row is still exactly as the retirement left it, so a manual disable in that window wins. It also covers the model having been re-enabled already, and the stamp having been cleared.

`SetEnabledIfConfirmed` is now `AutoRetireIfConfirmed`. It only ever disabled, and carrying an always-false bool hid that the write had acquired retirement semantics — which is how the disabled_manually mistake got in unnoticed.

**Test note:** the re-sighting test asserts the *discovery* case alongside the retirement one. "Does not revive" is only correct if the mechanism it shares with discovery still revives what it should, and asserting one without the other would let a fix that broke discovery pass.

One cosmetic gap I am flagging rather than expanding scope for: a traffic-retired model surfaces in the discovery-claim list, which is right (visible and actionable), but that list is worded around models that stopped being listed. A distinct claim label would be more accurate and is worth a follow-up.

6257 tests, lint clean, diff coverage 94.4%.

hugalafutro:
Fixed in aa29711. Correct, and it is the exact defect I fixed on the *other* side of the same window last round — I made the undo conditional and left the retirement writing by id alone. That asymmetry had no justification; I simply did not carry the reasoning across.

The write now locks the row and checks it still looks like the model the evidence was gathered against: enabled, not disabled by hand, not already retired.

`FOR UPDATE` is the part that makes the check meaningful rather than decorative. Without it the check and the commit are two separate moments and an operator write fits between them — which is precisely the bug being fixed, relocated a few lines. Holding the row from the check through the commit makes the whole decision atomic with respect to anything else touching that model, and it composes with the existing staging: `confirm` now runs inside a transaction that already owns the row.

A missing row is treated as an outcome rather than an error, since a model can be deleted between the decision and the write just as easily as it can be changed.

**What this deliberately does not do**, because I would rather state it than have it found: it does not distinguish evidence that *predates* an operator action. Strikes gathered before they enabled the model still read as a routable model here, and no row predicate can see that — it would need the streak to carry the row version it was gathered against.

I decided against adding one. That case resolves itself: if the model really is gone it refuses three more requests and is retired again with a fresh alert, which is the correct answer to an operator enabling a dead model. Adding a version column to detect it would trade a self-healing case for a permanently silent one, since a stale token would make the retirement stand down and the parked streak would stop it ever retrying.

Four DB tests, all against real PostgreSQL: an operator disable that must survive, a second retirement of an already-retired model, a deleted row, and the control that an untouched model still retires. Confirmed to fail against the unconditional write — the first two cases catch it, including `confirm` running when it should not have.

6262 tests, lint clean, diff coverage 93.7%.

Worth noting for the record: every finding from round 7 onward has been the same shape — a detached goroutine writing state that moved under it. Success cancelling a queued disable, a spent deadline, a replaced streak, an undo overwriting an operator, and now the retirement itself. They were fixed one at a time because each surfaced separately, but the row is now guarded on both writes and the streak on both reads, so the class is closed rather than the instances.

hugalafutro:
Fixed in 8961c71. Correct — the veto allowed one optional determiner between the preposition and the noun, so the wordier phrasings providers actually use walked straight past it. The qualifier may now run to a few words; it cannot cross punctuation, so it stays inside one clause, and it must still land on a whole capability noun.

**Two things worth recording, because one of them nearly shipped as a regression and the other means my first attempt at these tests was worthless.**

**1. The word boundary on the noun is load-bearing, not tidiness.** Without it `mode` matches the front of `model`, and Zens real retirement payload — `Model claude-sonnet-4 is not supported on the full model list` — becomes a capability refusal and stops retiring. The widened run is exactly what lets the pattern reach that far into the sentence, so the two changes are not separable: widening the qualifier without anchoring the noun trades your bug for a worse one. Two existing tests caught it, which is the best argument for keeping real captured payloads in the table.

**2. My first version of your five cases passed without testing anything.** I wrote them with `model-x`, and the tables helper falls back to an id that is absent from the body when it does not recognise one. An absent id makes the classifier decline to attribute the phrase at all, so all five went green against the *unfixed* pattern.

I only caught it because I ran them against the old regex to check they discriminated, which I now do as a habit — and it is the second time in this review that a case of mine passed for the wrong reason. So the trap is removed rather than avoided: tables whose every case must reach the wording checks now use `modelNamedInBody`, which fails loudly instead of falling back. That is the change I would most want kept from this round.

Confirmed the five cases fail against the previous pattern and pass against the new one, and all 84 classifier cases pass.

6267 tests, lint clean, diff coverage 92.7%.

hugalafutro:
No new inline findings this round — the most recent is the capability-veto one, fixed in 8961c71, and the summary was re-generated against that commit. So this is a reply to the three concerns the summary still lists rather than to anything new. All three are known; here is where each stands.

**1. Qualifiers longer than three words bypass the veto.** True, and I am keeping the bound — but not for the reason I would have given before testing it.

I assumed the bound was what stopped a long run reaching a distant capability noun in a genuine retirement. That was wrong: I probed `is no longer available and we recommend you migrate to a supported endpoint` at `{0,8}` and it still classifies as gone, because the run has to start immediately after `for|with|on|in` and `and` is not one. The preposition anchor does that work, not the count.

What the bound actually governs is preposition-anchored qualifiers, and there it does real work. At `{0,8}`:

```
Model gpt-4 is no longer available for anyone still using the deprecated completions endpoint
  {0,3} -> provider_model_gone      {0,8} -> provider_error
Model gpt-4 is no longer available for customers who have not migrated to the new plan
  {0,3} -> provider_model_gone      {0,8} -> provider_error
```

Both are retirements, and the wider bound turns both into missed ones. So raising the bound trades your false positive for a false negative, and a false negative here is the failure I would rather have: a stale model left routable costs a bad request, a false retirement disables a live one.

Three filler words already covers every phrasing either of us has produced from a real payload. Beyond that I would be guessing at wordings, and each widening step in this review has produced exactly one regression — the last one made `mode` match the front of `model` and broke Zens real retirement. Further phrasings should come from `/api/logs`, where the gateway now records them, rather than from speculation in review.

**2. Late successes and operator actions can lose to a queued retirement.** The remaining case is evidence gathered *before* an operator enabled the model, which no row predicate can see — it would need the streak to carry the row version it was gathered against. I decided against that in aa29711 and the reasoning stands: the case self-heals, because a model that really is gone refuses three more requests and is retired again with a fresh alert. A version token would replace a self-healing case with a permanently silent one, since a stale token makes the retirement stand down and the parked streak stops it retrying.

**3. A rollback can leave an asynchronously disabled custom group behind.** Narrowed to the post-commit rollback path only — the staged write closed the common case. When it does happen the group carries `auto_disabled_at` and surfaces as a discovery claim, which is the same treatment every auto-disabled group gets and is the designed operator path, not a silent hole. Restoring it automatically would be the first code in the codebase to re-enable a group without an operator, against the intent recorded in migration 062.

CI is green on every check. 6267 tests, lint clean, diff coverage 92.7%.

hugalafutro:
Revisiting the "operator actions lose to queued retirement" concern, because my previous reply was too quick to file it as a known residual — and re-reading the code rather than my own summary of it, part of it was reachable in a way I had not accounted for.

**Fixed in dce1848: the streak had no time bound at all.**

"Three consecutive refusals" counted refusals with no limit on how far apart they were. Two strikes from a provider incident in the morning and one in the afternoon combined into a retirement, on evidence whose halves had nothing to do with each other.

That is what makes the operator case reachable rather than theoretical, and it is the part I had been answering wrongly. I kept looking for a *race* — an operator write interleaving with a queued one — and arguing the window was tiny. The real path needs no race at all: old strikes are exactly the ones an operator has already responded to. They look at a failing model, enable it, and a count they cannot see finishes on the next refusal and turns it off again. No row predicate can catch that, because the evidence predates their action rather than competing with it.

So the fix is not another condition on the write. It is to stop carrying evidence that old: a strike arriving more than thirty minutes after the previous one starts the streak over.

Thirty minutes is picked against traffic rather than as a round number. Three requests to one model fall inside it on any gateway with load, while a model idle enough to fail three times over several hours no longer accrues a retirement at all — the right answer, since traffic-driven retirement should not be guessing from sparse evidence.

`Swap` is what keeps the reset safe under concurrency: only the caller that takes the stale timestamp sees it, so exactly one resets and the rest add.

The test drives the age through the streaks own timestamp, so nothing exists in production purely to make it testable. Confirmed to fail without the window — it auto-disables the model on hours-old evidence, with the retirement alert to match.

**What remains, precisely.** A genuinely concurrent operator enable — one landing inside the sub-second window between the threshold being crossed and the write committing — can still be overridden. That one is a true race rather than stale evidence, it needs the operator to act in that exact instant, and it self-corrects: the model is enabled again and, if it works, the next success clears the streak. I am not adding a row-version token for it, for the reason given in aa29711.

6268 tests, lint clean, diff coverage 93.3%.

hugalafutro:
Fixed in 467d186. Correct, and the part that makes it serious is that the suppressing event is the ordinary one, not an edge case.

`deriveStreamError` classifies the providers in-stream error, then overwrites `errorKind` with `client_disconnect` when the client hangs up. The retirement verdict was read from `errorKind`, so a client that received a "model is gone" chunk and closed the connection erased the strike that chunk had just earned — the evidence destroyed by the client reacting to it. Hanging up is exactly what a client does on seeing an error mid-stream, so a genuinely retired model could stay routable indefinitely: every refusal that actually reached a client would be filed as the clients disconnect and count for nothing.

The fix is to stop conflating two different questions. `errorKind` answers "how did this request end", and later causes legitimately overwrite earlier ones there — the log and the metrics want the disconnect. `upstreamKind` answers "what did the provider say", nothing overwrites it, and the retirement verdict is drawn from that.

I kept the disconnect as the recorded kind rather than giving the retirement precedence. Both are defensible, but changing which kind lands in `request_logs` would alter the 499 mapping and the metric labels for every disconnect-during-error, which is a behaviour change well outside what this finding asks for. The verdict was the thing that was wrong, so the verdict is the thing that changed.

Covered at both levels deliberately: the verdict table pins the rule, and `deriveStreamError` is tested end to end for the wiring — the table alone would pass happily with `upstreamKind` never being set anywhere. Both confirmed to fail against the previous behaviour.

Also noting the summary no longer lists the queued-retirement or capability-veto concerns, so I am treating those as settled from your side too.

6271 tests, lint clean, diff coverage 93.1%.

hugalafutro:
Both fixed in b95a02f, and the first one includes a correction to what I told you last round.

**1. Concurrent reset loses strikes.** Correct. I wrote that a reset could cost at most one strike; that was wrong. `Store(1)` discards however many increments raced it, so a burst of refusals at the window boundary can end on a count of one and never reach the threshold — a dead model left routable by the very traffic proving it is dead.

The mistake was treating "has the window lapsed" and "apply that decision" as two operations. They are one, and two atomics cannot express it. The count and the last-strike time now move together under a mutex, held for a comparison and an addition on a per-model struct that is only contended while that one model is failing.

The test runs every strike at the boundary so the reset and increment branches race on each iteration rather than by luck, 200 iterations under `-race`. Against the split version it fails with `streak = 2, want 3`.

**2. Stale revert defeats a fresh retirement.** Also correct, and the sequence is nastier than it first reads because the replacement retirement standing down is the *right* behaviour at that moment — the disable it wanted is already committed. It is the older revert that is wrong, and it then leaves the replacement streak parked above the threshold where it will never disable anything again. So the model ends up enabled with three fresh refusals against it and no mechanism left to act on them.

The revert now defers to what the model is saying now rather than to the success that scheduled it: if a current streak has reached the threshold, the retirement stands.

That is the same principle as the `cancelled` tombstone, applied one step further out. The tombstone asks "has this decision been superseded"; this asks the same question of the *undo*, which was the one path still assuming its own evidence was the latest.

**On the test itself:** the first version recursed, because the mock hook fires on every staged write and the replacement retirement is itself a staged write. Worth mentioning because it is the same shape as the bug — something scheduled by an earlier event running again in a context that has moved on.

Both confirmed to fail against the previous code.

6273 tests, lint clean, diff coverage 93.4%.

hugalafutro:
Fixed in 4abef83. Correct, and it is my own comment being read back accurately — I wrote that a stream which emitted content "will normally set at least one" of completion tokens or TTFT. Normally is not always, and both signals are optional in ways that co-occur.

A provider that omits the usage chunk, on a gateway with the TTFT probe switched off, streams a perfectly good answer that reads as having produced nothing. The success never clears the strike streak, and later refusals retire a model whose failures were never consecutive. Omitting usage is common and the probe is a setting, so that is an ordinary configuration rather than a corner.

The stream now records that content actually flowed, set where the content is observed and nowhere else. The other two stay as corroboration for paths that never parse deltas.

**The bar deliberately stays positive evidence rather than absence of an error**, because the opposite failure is just as real and was an earlier finding in this review: a stream can open, emit nothing and end without recording anything, and crediting that would clear a retirement streak on the strength of an empty response. So the observer tests pin both directions — content and reasoning deltas count; empty deltas, role-only opening chunks and error chunks do not. That last one matters most: a refusal must never read as the model having answered.

Native Anthropic passthrough never parses deltas, so its terminal `message_stop` stands in for the same fact.

Tested at both levels on purpose. The unit case would pass with the flag never being set anywhere, so the observer is tested directly against real chunk payloads. Confirmed to fail without the change.

6280 tests, lint clean, full proxy package green under `-race`, diff coverage 93.5%.

PR #610, opened 2026-08-03
fix(discrepancies): re-read the Models badge after every discovery write

hugalafutro:
## What

The Models nav badge is a 60s poll on its own query key, and nothing that changes the claim set touched it. Dismissing every claim and closing the modal left the badge asserting a count the server no longer had, until the next poll landed. Confirmed against prod MH1's access log:

```
14:50:25.346-.359  POST /api/discovery/dismiss  x7  (all 200)   <- "Dismiss all"
14:50:25.393       GET  /api/discovery/status  3472 bytes       <- modal's own refresh
   ... 11s gap, no requests ...                                 <- modal closed, badge still lit
14:50:36.620       GET  /api/discovery/status  3472 bytes       <- reopened (?review=1)
14:50:36.672       POST /api/discovery/changes/ack             <- journal expanded
14:50:36.703       GET  /api/discovery/status  86 bytes         <- badge finally cleared
```

## How

A shared `useRefreshDiscoveryBadge()` owns the invalidation, so the `exact: true` that keeps the modal's `?review=1` query (and its server-side last-reviewed stamp) off a click is stated once instead of at every call site.

**Every surface that can move `claim_count` now calls it**, which took three passes to get complete:

| Surface | What moves the count |
|---|---|
| Discrepancy modal | all three dismiss paths |
| `useDiscoveryRetest` | so a retest refreshes the badge from the Providers page too |
| Providers page | discover, discover-all, provider delete, disabled-model bulk delete |
| Models page | discover, single delete, bulk delete, **and the enable/disable toggle** |
| Settings | the second "Discover all" |
| EditProviderModal | `listClaimRows` joins on `p.enabled`, so switching a provider off takes every claim it owns |
| FailoverGroups + CreateGroupModal | an auto-disabled group IS a counted claim, so re-enable / delete / sync all move it |
| AddProviderModal | the scan after a create re-syncs failover, and fresh members can lift a group back over the routable floor |

Two subtleties worth reading the comments for:

- **Toggling reclassifies rather than removes.** `buildProviderClaims` puts an enabled model in Suspect, and Suspect is not counted, so enabling a Gone model drops the count without deleting anything.
- **`useDisableModel` is deliberately excluded.** It only ever acts on a model from `enabledModels`, and an enabled model is either absent from the claim set or Suspect, so `claim_count` cannot move. `UpdateModel` does no failover resync either. Recorded in the file so it is not "fixed" later.

**Refreshes moved from `onSuccess` to `onSettled`, and the lists move with the badge.** A rejected write can still have landed (the response is what was lost), and a discovery run that errors partway has still upserted whatever it reached. Re-reading only the badge would leave the page contradicting itself: the count dropping for a provider whose card is still on screen.

Seeding the badge from `refresh()`'s return value instead would save a round trip but is unsound: that value is returned even when a newer refresh has superseded it, so it can write a known-stale count into the badge. Rationale is recorded in the hook.

## Also: Escape stopped closing modals

Found while verifying the above in a real browser. Escape was handled on the dialog node, so it only saw keys originating inside that subtree. Dismissing a claim unmounts the button that had focus, focus falls to `<body>`, and the modal goes permanently deaf to Escape. Reproduced live: `document.activeElement = <BODY>`, Escape does nothing.

Now handled on the document, which does not depend on focus semantics. Restoring focus would also have fixed Tab order, but browsers disagree about whether removing a focused element fires `focusout`, so that fix would work in some and not others. A module-level stack keeps nesting correct: only the topmost dialog reacts, so a nested confirm closes before its opener. Affects every modal in the app.

## Tests

**20 new tests.** The badge ones mount each page inside the real `<Layout>` and read the real badge, against a status endpoint whose `claim_count` actually falls as writes land: a fixed payload would be right by accident whether or not anything re-read it. Several drive paths that **report failure**, since that is what the `onSettled` placement buys.

Every one was verified to fail with the fix removed, and only those (neutralising `useRefreshDiscoveryBadge` fails the 9 badge tests and nothing else; removing the `providerId` guard fails only the guard test; reverting `Modal.tsx` fails only the 4 Escape tests).

Full frontend suite green (294 files / 5657 tests), `tsc -b` and `pnpm lint` clean.

## Verified live

Rebuilt the dev container and drove it with a real browser against 114 real claims. The discriminator is **time**: the poll is 60s and a reload restarts it, so a change within 10s can only be an invalidation.

| Check | Result |
|---|---|
| Dismiss a claim, close the modal, read the badge | 110 → 109 in **5ms** |
| Enable a Gone model from the Models page | 109 → 108 in **209ms** |
| Badge agrees with `/api/discovery/status` | both 108 |
| Disable a provider drops its counted claims | 108 → 100 → 108 |
| Escape after a dismissal | closes (was: nothing) |

That fourth row is a better check than it looks: xAI owns 9 claim rows but only **8** are counted, the 9th being Stale, and the count dropped by exactly 8.

Dev was restored to exactly its prior state afterwards (`claim_count` back to 114).

### DIFF
--- a/web/src/components/Layout.tsx
+++ b/web/src/components/Layout.tsx
@@ -40,6 +40,7 @@ import {
 import { useGitHubVersion } from "../hooks/useGitHubVersion";
 import { useIdleLogout } from "../hooks/useIdleLogout";
 import { useReadOnly } from "../hooks/useReadOnly";
+import { useRefreshDiscoveryBadge } from "../hooks/useRefreshDiscoveryBadge";
 import i18next, { LANGUAGE_STORAGE_KEY } from "../i18n";
 import { useDiscoveryRetest } from "../pages/Providers/useDiscoveryRetest";
 import { CollapsibleToggle, useCollapsible } from "./CollapsibleToggle";
@@ -756,6 +757,13 @@ export function Layout({ children }: LayoutProps) {
 	const { toast } = useToast();
 	const readOnly = useReadOnly();
 
+	// Requirement 5 keeps the ?review=1 stamp off the 60s timer; the `exact: true`
+	// inside this hook is what keeps it off a click. See useRefreshDiscoveryBadge.
+	//
+	// Retests do NOT call it from here: useDiscoveryRetest owns that, so a retest
+	// run from the Providers page refreshes the badge too.
+	const refreshBadge = useRefreshDiscoveryBadge();
+
 	// The retest response's own diff is deliberately discarded. It describes what
 	// THAT run changed, which is empty when the model is still missing, and
 	// reading that emptiness as "fixed" is the original defect. Truth comes from
@@ -928,9 +936,15 @@ export function Layout({ children }: LayoutProps) {
 					}),
 					"error",
 				);
+			} finally {
+				// `finally`, not the success path: a request that rejects can still
+				// have landed — the response is what was lost — and the rows correctly
+				// stay actionable either way. Re-reading the badge is how it learns
+				// which of the two happened.
+				refreshBadge();
 			}
 		},
-		[dismissClaim, refresh, toast, t],
+		[dismissClaim, refresh, refreshBadge, toast, t],
 	);
 
 	const onDismissEverything = useCallback(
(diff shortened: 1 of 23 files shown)


hugalafutro:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #680, opened 2026-08-15
feat(auth): sliding session expiry (3 days idle, 30-day cap)

hugalafutro:
Follow-up to #678. A session now lives **3 days without use**; every use slides that window forward, up to an **absolute cap of 30 days from login**. Before, the 3 days ran from login regardless of activity.

## How
- `webauthn.SessionManager.Authenticate` is the new validation entry point (`TokenUser` wraps it). On the same 5-minute throttle as the last-seen stamp it slides `expires_at` to `now+AuthTokenTTL`, clamped to `created_at+AuthTokenMaxLifetime`; only ever forward, and only when the gain is ≥1 minute (the first request after login finds no stamp and would otherwise re-issue for milliseconds). A session past the cap is rejected on validation whatever its stored expiry says. Extension write failures never fail authentication (mirrors the last-seen stamp).
- `SessionStore.ExtendSession` added; implemented for Postgres (`Repository`) and Front Desk's SQLite (`WebAuthnStore`).
- Browsers enforce cookie MaxAge on their own, so both cookie gates re-issue the cookie pair with the new lifetime when `Authenticate` reports a slide: the dashboard `AuthMiddleware` (`resolveCredentials` now returns a refresh hint) and `adminauth.RequireAdminOrSession` (Front Desk; gains a `cookieSecure` param). New `authcookie.Jar.RefreshSession` keeps the CSRF cookie's value and moves it with the session cookie. SSE re-checks ignore the signal (headers already sent).
- TTL test pins both constants (72h idle, 30d cap) under independent security ceilings (7d, 90d) and asserts idle ≤ cap.
- Copy: Tab timeout hint in all 29 dashboard + 11 Front Desk locales now says "3 days without use, and after 30 days at most"; README (2 mentions), wiki Security + Configuration updated.

## Tests
- webauthn (Postgres): slides on touch, no slide inside the throttle, first request after login doesn't slide, clamps to the cap and stops at it, rejects past the cap, never shortens a legacy longer expiry, extend failure doesn't fail auth, TokenUser contract.
- authcookie: RefreshSession keeps values / moves MaxAge / mints CSRF when missing.
- api AuthMiddleware: re-issues cookie pair on slide, writes nothing otherwise. adminauth RequireAdminOrSession: same, plus honours `cookieSecure=never`.
- frontdesk SQLite: ExtendSession moves only expires_at, missing row quiet, refused write surfaces + auth still passes; end-to-end slide through the manager.
- Mutation-checked the extension write and both re-issue sites. Diff coverage 90.7%; golangci-lint clean; MH/FD `vitest related` on en.json green.

Live-verified on the dev FD: fresh request → no Set-Cookie; aged session (last seen 1h ago) → both cookies re-issued with Max-Age 259199 and `expires_at` moved to now+3d; immediate follow-up → no Set-Cookie.

Sessions minted before deploy keep their stamped expiry until first use slides them (or the cap catches them).


hugalafutro:
Opus review round applied in bca3f99f:
- SSE heartbeats (dashboard reauth loop, FD `revalidate`) now go through a new non-sliding `SessionManager.Verify`; only real requests slide (mutation-checked at both call sites).
- `RefreshSession` echoes the request CSRF value only when it has the minted shape (43 base64url chars), sets `Cache-Control: no-store`, ignores non-positive MaxAge, shares `writePair` with `SetSession`.
- Zero `created_at` is logged (both stores declare it NOT NULL); listing/cleanup honour the cap by construction (comment on `Authenticate`).
- New tests: non-admin session cookie refused with no cookies re-issued; Verify writes nothing; slid session still listed; mint path stores created_at; foreign CSRF replaced; non-positive MaxAge writes nothing.

Known and accepted: a GET whose response lands after a logout response can re-set the cookie pair client-side; the server row is already gone, so the next request 401s and the SPA's `clearAuth()` recovers. Narrow, self-correcting; not worth skipping refresh on unsafe methods for.

PR #796, opened 2026-08-27
fix(http): bound every JSON request body

hugalafutro:
## What

Every HTTP handler on the three admin surfaces (`internal/api`, `internal/adminauth`,
`internal/frontdesk`) decoded its request body with a bare
`json.NewDecoder(r.Body).Decode(&x)` and no ceiling of its own.

On the main server that was not unbounded: the router already caps every request at
`MAX_REQUEST_SIZE` (50 MB by default). But 50 MB is sized for the proxy's multimodal uploads, and
it is the wrong bound for a login body. **The Front Desk binary mounts no global cap at all**, so
its handlers really did read as much as a client was willing to send, including on
`POST /api/pair`, the Bellhop pairing exchange, which is unauthenticated by design with only the
per-IP limiter in front of it.

The other routes that answer before authentication: `POST /api/auth/admin-exchange` on both
binaries, and the password, TOTP and WebAuthn ceremonies in `internal/adminauth`.

## How

`httpx.DecodeJSON` binds the body with `http.MaxBytesReader` before decoding, answers **413** past
the limit and **400** for a body that is malformed or carries anything after its one JSON value.
The bound counts bytes rather than trusting `Content-Length`.

Draining what follows the value is what makes the ceiling real rather than nominal: `json.Decoder`
stops the moment it has a complete value, so without the drain a tiny object with a huge tail
decodes fine and never trips the limit. A trailing newline is not a tail.

- Default `httpx.MaxJSONBody` = 1 MiB. The largest legitimate payload on any route using it is a
  full 10 000-ID bulk model delete at roughly 400 KiB, pinned by a test that decodes the real
  worst case through the real helper.
- `decodeJSONLimit` carries the per-endpoint ceilings: config-sync import (8 MiB) and fleet
  announce (1 KiB), both unchanged, plus a new named ceiling for the fleet quota receiver, which
  previously sat at exactly the Front Desk read ceiling that feeds it.
- `DecodeJSONOptional` tolerates an absent body only. A body that arrived and could not be read is
  a 400, not a silent fall back to the default: on `DELETE /api/logs/app` that was the difference
  between clearing an hour of logs and clearing all of them.
- Decode failures are logged as a fixed phrase plus a byte offset, never the decoder error, whose
  message quotes the offending byte. These routes answer before authentication and this gateway
  does not log request content.

## Guard

`internal/httpx/decodeguard_test.go` walks the repository and fails on any
`json.NewDecoder(r.Body)` outside the package that implements the bounded read, skipping comments
and lines that bind the reader in place. Verified to fire on a probe handler planted in `cmd/`.
Its doc states what it does not catch: it matches one call shape, and it says nothing about
whether a given limit is a sensible size.

## Also fixed, found in review

A member's display name had no length cap, and the primary's name rides in every fleet announce
against a 1 KiB ceiling, so a long enough name broke announces fleet-wide and near-silently (the
poller logs a failed announce at debug level only). Capped at 200 characters on both write paths,
rejected rather than truncated, with a test tying the cap to the announce budget.

## Verified

- `go test` on the four touched packages, `golangci-lint run`, `make size-check`, `go build ./...`
- Live on the rebuilt dev stack (app :8081, Front Desk :8090):

  | surface | case | result |
  |---|---|---|
  | MH `/api/auth/admin-exchange` | oversized / malformed / valid | 413 / 400 / 200 |
  | MH `/api/auth/admin-exchange` | valid value + trailing second value | 400 |
  | MH `/api/auth/admin-exchange` | valid value + trailing newline | 200 |
  | MH `/api/providers` (authed) | oversized | 413 |
  | MH `/api/fleet/announce` | 2 KiB body | 413 (1 KiB ceiling intact) |
  | MH `/api/fleet/announce` | tiny value + tail past the ceiling | 413 |
  | MH `DELETE /api/logs/app` | truncated `older_than` | 400, nothing cleared |
  | FD `/api/pair` (unauth) | oversized / malformed / wrong code | 413 / 400 / 401 |
  | FD `POST /api/members` | 250-character name | 400 |

- No client keys off the old error text: nothing in `web/src`, `frontdesk/web/src`, `web-shared`
  or `android/` matches it, and both machine-to-machine consumers (the announce poller, config
  sync, quota distribution) funnel 400 and 413 into the same arm.

## Context

Raised by a Strix AI pentest run. Four other recommendations from the same batch were checked and
rejected: `COOKIE_SECURE` is already `always` by default, blocking `fc00::/7` in `netguard` would
break the internal IPv6 destinations that package exists to reach, the fair-share limiter's 24 h
divisor TTL is a deliberate tradeoff rather than a fail-open, and routing the version check
through `SafeDialer` guards a compile-time constant URL. The Traefik config endpoint was already
gated in #747.

### DIFF
--- a/internal/adminauth/bodylimit_test.go
+++ b/internal/adminauth/bodylimit_test.go
@@ -1,0 +1,52 @@
+package adminauth
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/hugalafutro/model-hotel/internal/authcookie"
+	"github.com/hugalafutro/model-hotel/internal/httpx"
+)
+
+// TestTokenExchange_OversizedBody_413 covers this package's defining property:
+// every ceremony here (token exchange, password login, TOTP, WebAuthn) is
+// reachable before authentication, so their request bodies are the ones an
+// anonymous caller fully controls. The body is valid JSON with a valid token,
+// so only the size limit can reject it.
+func TestTokenExchange_OversizedBody_413(t *testing.T) {
+	sessionMgr := newTestSessionManager(t)
+	adminMgr := &mockAdminAuth{validateFn: func(token string) bool { return token == "sekrit" }}
+	h := TokenExchange(adminMgr, sessionMgr, nil, authcookie.FrontDesk, "never", nil)
+
+	body := `{"admin_token":"sekrit","padding":"` + strings.Repeat("a", httpx.MaxJSONBody+1) + `"}`
+	rec := httptest.NewRecorder()
+	h(rec, httptest.NewRequest(http.MethodPost, "/api/auth/admin-exchange", strings.NewReader(body)))
+
+	if rec.Code != http.StatusRequestEntityTooLarge {
+		t.Fatalf("status = %d, want 413; body=%q", rec.Code, rec.Body.String())
+	}
+	if len(rec.Result().Cookies()) != 0 {
+		t.Errorf("a rejected exchange must mint no cookie, got %+v", rec.Result().Cookies())
+	}
+}
+
+// TestTokenExchange_MissingToken_400 pins the split of the old combined
+// decode-or-empty-token condition: an empty token is still a 400, not a 413 and
+// not a session.
+func TestTokenExchange_MissingToken_400(t *testing.T) {
+	sessionMgr := newTestSessionManager(t)
+	adminMgr := &mockAdminAuth{validateFn: func(string) bool { return true }}
+	h := TokenExchange(adminMgr, sessionMgr, nil, authcookie.FrontDesk, "never", nil)
+
+	rec := httptest.NewRecorder()
+	h(rec, httptest.NewRequest(http.MethodPost, "/api/auth/admin-exchange", strings.NewReader(`{}`)))
+
+	if rec.Code != http.StatusBadRequest {
+		t.Fatalf("status = %d, want 400; body=%q", rec.Code, rec.Body.String())
+	}
+	if len(rec.Result().Cookies()) != 0 {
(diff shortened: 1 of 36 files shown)


hugalafutro:
Confirmed and fixed in 716ae7d4. The error message was the honest half; the implementation was not.

`validMemberName` now counts with `utf8.RuneCountInString`. That moves the cap into a different unit than the constraint behind it, which is worth stating: the cap exists because the primary's name rides in every fleet announce, and the receiving handler bounds that body at 1 KiB. A character is up to 4 bytes of UTF-8 and up to 6 once JSON escapes it, so counting characters at 200 would have allowed a 1200-byte name and broken announces fleet-wide, near-silently (the poller logs a failed announce at debug level only). The cap is therefore 128, whose worst case is 768 bytes.

Rather than trust that arithmetic, `TestMemberNameFitsAnnounceBudget` marshals a real worst-case announce (128 four-byte characters beside a full-length Front Desk id) and asserts it fits. Mutation-checked: raising the cap to 256 fails it at 1137 bytes, and reverting the counter to `len()` fails `TestMemberNameCapCountsCharactersNotBytes`.

Your finding also pointed at the same bug one file over, on the endpoint this PR exists for. `CreatePairedDevice` truncated with `label[:maxDeviceLabelLen]`, slicing bytes: a 150-character Chinese label came out as 33 characters plus a broken half of a fourth, which renders as a replacement glyph everywhere it is shown. That label arrives in the unauthenticated pairing request, so its content is entirely the caller's. Truncation now cuts on character boundaries, with a regression test that fails on the stored `\xe7` tail under the old code.

Verified on a rebuilt dev Front Desk: a 128-character Chinese name passes name validation and falls through to the URL scheme check, and 129 is refused with "name must be at most 128 characters".

PR #799, opened 2026-08-27
fix: keep the FD login token out of the structured log, meter every passthrough path

hugalafutro:
Two findings from the GLM 5.2 Strix run, plus the paths review found still uncovered after the first attempt.

## 1. Front Desk logged its generated admin token as a structured attribute

```go
debuglog.Info("frontdesk: generated Front Desk login token", "token", adminMgr.Token())
```

`debuglog`'s handler fans out to every configured channel, so on first boot the plaintext login credential reached the JSON stdout handler — where it becomes an **indexed, queryable field** — and the **OTLP exporter**, which ships it off the host into the operator's log store and retains it under that store's policy. The gateway has always printed its own token straight to stdout for exactly this reason (`printAdminTokenBoxStdout`); Front Desk was the asymmetry.

Front Desk now uses `announceGeneratedToken`, writing to stdout and logging only the token-free fact that one was generated.

**Stated precisely**, because review rightly pushed back on the first wording: in a container, stdout *is* the log stream, so the token still reaches whatever collects container output. What this removes is the structured indexed attribute and the OTLP hop into a remote store. That is a real reduction in reach and retention, not a guarantee the value never lands in a log file.

The write error is also no longer discarded. That `Fprintf` is the **only** copy the operator gets — the token is persisted as a hash — so a silently lost write means deleting the token file and restarting to recover. The token must not be logged; the failure must be.

## 2. Pass-through responses without a usage block were never metered

`extractPassthroughUsage` returns `(0,0)` when a provider reports no usage, and the debit was guarded on one of them being non-zero. So the request was charged against neither the TPM budget nor `tokens_used`.

**The first attempt fixed one branch and left the class** — the same criticism its own message quoted, earned again. Review caught two families still unmetered:

**Text-to-speech never reached the fixed branch.** `/v1/audio/speech` answers `audio/mpeg`, which is neither JSON nor SSE, so it routes to `serveStreamedPassthrough`. There the tail buffer carrying usage is only allocated `if isSSE`, so for binary audio the report is *structurally always absent* and the old guard could never be true. TTS was unmetered before and after that commit. Its test asserted otherwise by constructing an `application/json` response TTS does not return.

**Speech-to-text and image edits were a no-op for a different reason.** `ingestMultipartRequest` never assigned `promptTextBytes`, so the estimate computed from zero — the same silent-zero failure as the chat-only sizer, one family over. The asymmetry was visible in the result: `/v1/images/generations` charged an estimate while `/v1/images/edits`, same model and same `prompt` field, charged nothing purely because the prompt arrived as form data.

Both paths now charge through one `chargePassthroughUsage` helper, and `multipartPromptTextBytes` sizes the text form fields while skipping the upload — the same rule every other sizer here applies to blobs.

## 3. The estimate is gated on delivery

Charging whenever the provider reported nothing meant a 200 carrying `{"data":[]}` — an aggregator in front of a retired model, precisely what `passthroughAnswered` exists to detect — billed the caller a full prompt estimate for an empty answer, on every request. `estimateMissingUsage` already states the rule this restores: nothing is estimated when nothing was delivered.

Reported figures always win; the estimate fills only a total absence, covers the prompt alone (a response of vectors, base64 or audio is not text), and reaches `recordTokenUsage` without entering the request log, which continues to report only what the provider measured.

## Verification

- Both new guards mutation-verified: reverting the streamed-path charge fails the TTS test, and removing the delivery gate fails with `charged 1 times for an empty answer, want 0`.
- The mislabelled TTS test is now an honest image-generation test, beside a real binary-response one that exercises the route TTS actually takes.
- `go test` on both touched packages (1,561 pass), `golangci-lint` clean, `make size-check` clean, `-race` clean.

**Known limit, stated rather than glossed:** `TestGeneratedTokenNeverReachesTheStructuredLogger` covers the helper, not `main()`. Re-adding the original `debuglog.Info(..., "token", ...)` call to `main()` would leave it green. `cmd/` is coverage-excluded and a `main()`-level test is awkward, so the guard is narrower than ideal.

### DIFF
--- a/cmd/frontdesk/main.go
+++ b/cmd/frontdesk/main.go
@@ -10,6 +10,8 @@ package main
 import (
 	"context"
 	"errors"
+	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"net/url"
@@ -110,8 +112,9 @@ func main() {
 		debuglog.Fatal("frontdesk: failed to initialize admin token", "error", err)
 	}
 	if isNew {
-		// Printed once so the operator can capture the generated UI login token.
-		debuglog.Info("frontdesk: generated Front Desk login token", "token", adminMgr.Token())
+		// Straight to stdout, never through the structured logger: see
+		// announceGeneratedToken.
+		announceGeneratedToken(os.Stdout, adminMgr.Token())
 	}
 
 	bus := events.NewBus()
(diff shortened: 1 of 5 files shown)


hugalafutro:
Confirmed and fixed in 8e81c33f. You were right, and it was an overcharge I introduced in this branch rather than a pre-existing one.

`multipartPromptTextBytes` skipped the upload and the `model` field and counted everything else, but everything else on these forms is configuration: `language`, `temperature`, `response_format`, `size`, `n`, `voice`, `timestamp_granularities`. A transcription sending only options and a file was billed for the text of its own settings.

It is now an allowlist of the fields that carry prompt text, which is just `prompt`: an image edit's prompt is the same string `/v1/images/generations` sends as JSON, and a transcription's prompt is the context hint the model conditions on. An allowlist rather than the endpoint-aware denylist your comment suggested, for one reason worth stating: a denylist gets it wrong by default every time a provider adds a parameter, and "silently becomes billable" is the wrong direction for a billing decision to fail in. If a future family gains a genuine second prompt-bearing field, adding it to the map is a deliberate one-line change with a test.

`TestMultipartPromptTextBytes_MetadataOnlyFormChargesNothing` covers exactly your repro: a form of nothing but metadata and a file must cost zero. Mutation-verified — restoring the previous denylist makes it charge 29 bytes of pure configuration.

PR #806, opened 2026-08-28
fix(proxy): forward SSE frames the gateway has no type for

hugalafutro:
## The bug

Every SSE `data:` payload is unmarshalled into a `streamChunk` struct, and a frame that failed was dropped — *"Drop invalid/truncated JSON instead of forwarding broken bytes"*. Sound intent, too narrow a struct: `delta.content` is typed `*string` and tool-call `arguments` is typed `string`, so a provider sending either in a wider but perfectly valid shape failed the unmarshal and had its answer discarded as though the bytes were corrupt.

The tool-call case is the sharp one, because the loss isn't uniform. `finish_reason` rides a *separate* frame that parses fine, so it survives. Reproduced against the dev stack with `arguments` as an object:

```
UPSTREAM SENDS:
  data: {...tool_calls:[{...function:{"name":"get_weather","arguments":{"city":"Prague"}}}]}
  data: {...finish_reason:"tool_calls"}

CLIENT RECEIVES:
  data: {...finish_reason:"tool_calls"}
  data: [DONE]
```

A completion announcing a tool call with no tool call in it. Most SDKs will throw or wait forever for arguments that never arrive.

Syntactically valid JSON objects are now forwarded; only genuinely malformed bytes are still dropped, which is what that guard was for.

## What the forwarded frame still has to obey

The frame skips the typed-parse branch, so everything living inside that branch had to be re-established. Review found three that weren't, one of them a regression this PR introduced:

**`strip_reasoning` is a per-virtual-key contract** and its transform is inside that branch, so a forwarded frame delivered reasoning to a caller configured never to receive any. Dropping the frame had been satisfying that guarantee by accident. Confirmed byte-for-byte before the fix. `computeStripReasoning` now runs on this path too — it works because `parseChunkPayload` is map-based and doesn't care about field types — and `deltaHasContent` no longer judges only the plain-string form of `content`, which would have turned a content-as-parts delta into an empty keep-alive and lost the text a second way.

**Only top-level JSON objects are forwarded.** `json.Valid` also admits `123`, `"text"`, `true` and arrays; a client decodes every `data:` event into a chunk struct, so relaying one of those converts a one-frame loss into a decode exception that kills the whole stream. That gate is also what keeps `carriesErrorObject` meaningful, since it reports false for anything that isn't an object — so a key-shaped token quoted inside an array-shaped error frame no longer escapes the key-shape mask.

**Only frames that carry output are forwarded, and only their output is metered.** `deliveredBytes` is the sole input to `estimateMissingUsage`, which charges quota and TPM. An earlier revision counted the whole JSON envelope, which over-charged wildly and — worse — gave a stream that delivered nothing but an error a non-zero count, unlocking a full *prompt*-size estimate for a request the provider never billed. `unmodelledDeliveredBytes` now sizes only the delta members that hold output, and a frame measuring zero is not forwarded at all: an error envelope the typed path cannot read (Ollama's bare `{"error":"model not found"}`) behaves exactly as it did before this PR. That also keeps a failing provider receiving the circuit-breaker charge it should — a zero-output frame counted as delivery was letting it escape one.

**The strip path sizes what survived the strip**, so a caller is never billed for reasoning deliberately withheld from them, and a keep-alive costs nothing.

**Forwarded frames are not counted as unparsed.** That counter makes the breaker verdict withhold judgement because a frame's contents are unknown — but reaching the forward path means output was measured in them. Left in, a provider whose normal shape lands here could never record a breaker success, and `consecutiveFails` only resets on success, so five failures spread over any span would open its circuit.

**Masking is repeated here**, since it also lives in the skipped branch: exact-key always, key-shape when the frame carries an error member, matching the typed branch's policy. Without it a forwarded frame reaches the caller carrying the provider credential — a test pins this with a deliberately non-key-shaped credential.

**The warning is once per stream.** Forwarding makes the unmodelled shape a normal operating mode for some providers, so a per-frame Warn would ship thousands of lines per request to Loki/OTLP.

## Verification

Every fix is mutation-tested, and two of those mutations initially produced *no* signal — one broke compilation, one had no test behind it. Both were chased down rather than counted as passes:

| mutation | result |
|---|---|
| drop unmodelled frames again | 3 tests fail |
| forward malformed bytes | original guard fails |
| skip masking | credential reaches the caller |
| bypass `strip_reasoning` | reasoning reaches the caller |
| allow non-object payloads | 5 subtests fail |
| forward frames carrying no output | error-only frame is forwarded and billed |
| count the envelope instead of the output | 9 subtests fail |
| remove the key-shape mask | key-shaped token reaches the caller |
| narrow `deltaHasContent` back to strings | content-as-parts becomes an empty keep-alive |

Two mutations initially produced *no* signal and were chased rather than counted as passes: one broke compilation, one had no test behind it. A third revealed a test passing vacuously — the key-shape test used an error-only frame, which this revision drops, so the mask was never reached; it now uses a frame carrying both an error member and real output.

The metering test asserts through the estimate's log line, not `logData` — the estimate runs after `logData.tokensCompletion` is assigned and its result goes to the meter and quota, never back onto the request-log row. The first version asserted `logData` and failed against a correct implementation.

Prod app logs show zero occurrences of the drop across all four fleet members, so this was latent rather than active.

### DIFF
--- a/internal/proxy/proxy_stream_response.go
+++ b/internal/proxy/proxy_stream_response.go
@@ -419,6 +419,36 @@ func (h *Handler) handleDataChunk(sink *streamSink, st *streamState, ev sseEvent
 		}
 	}
 	if !written && !jsonValid {
+		// Valid JSON this gateway has no Go type for is still the provider's
+		// answer, and must reach the caller.
+		//
+		// streamChunk types delta.content as *string and tool-call arguments as
+		// string. A provider sending either in a wider — but perfectly valid —
+		// shape fails the typed unmarshal above, and used to be discarded here
+		// as though the bytes were corrupt. The caller silently lost real
+		// output; on a tool call it lost the call while finish_reason, which
+		// rides a separate parseable frame, survived to announce it.
+		//
+		// The masking that every other frame gets lives inside the typed branch
+		// above, so it is applied here or not at all: exact-key always, and the
+		// key-shape pass when the frame carries an error member, matching the
+		// policy there. carriesErrorObject rather than a second opinion about
+		// what an error frame is.
+		if json.Valid([]byte(payload)) {
+			masked := st.masker.maskExact([]byte(payload))
+			if carriesErrorObject(masked) {
+				masked = maskKeyShapedTokens(masked)
+			}
+			// Still counted: forwarding the bytes does not tell us whether they
+			// carried output, so the end-of-stream verdict keeps withholding
+			// its judgement rather than charging the provider for an emptiness
+			// it cannot vouch for.
+			st.unparsedChunks++
+			debuglog.Warn("proxy: forwarding an SSE frame the gateway cannot model",
+				"model", logData.modelID, "provider", logData.providerName,
+				"chunk_number", chunkCount, "payload_bytes", len(payload))
+			return !st.emitData(sink, masked, "unmodelled frame", chunkCount, logData)
+		}
 		// Drop invalid/truncated JSON instead of forwarding broken bytes.
 		preview := payload
 		if len(preview) > 80 {
--- a/internal/proxy/proxy_stream_response.go
+++ b/internal/proxy/proxy_stream_response.go
@@ -419,35 +419,8 @@ func (h *Handler) handleDataChunk(sink *streamSink, st *streamState, ev sseEvent
 		}
 	}
 	if !written && !jsonValid {
-		// Valid JSON this gateway has no Go type for is still the provider's
-		// answer, and must reach the caller.
-		//
-		// streamChunk types delta.content as *string and tool-call arguments as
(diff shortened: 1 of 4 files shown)


hugalafutro:
Closing in favour of a narrower fix. This built a parallel path around the typed parser, which meant re-deriving by hand everything the typed path feeds — delivered-bytes accounting, the breaker verdict, the retirement verdict, the strip_reasoning contract. Two review rounds found breakage in a different one of those each time, and the remaining fix list was a fourth hand-derivation of work the typed path already does exactly.

The actual bug is that streamChunk is too narrow. Widening it so these frames parse normally removes the second path entirely: observers, masking, strip, metering, breaker and retirement then work by construction. Replacement PR to follow.

PR #807, opened 2026-08-28
fix(proxy): widen streamChunk to the shapes providers actually send

hugalafutro:
Replaces #806, which took the wrong approach. Same bug, much smaller change.

## The bug

Every SSE `data:` payload is unmarshalled into `streamChunk`, and a frame that failed was dropped. But the struct typed `delta.content` as `*string`, tool-call `arguments` as `string`, and the error member as an object — so a provider sending any of them in a **wider but perfectly valid** shape had its answer discarded as though the bytes were corrupt.

The tool-call case is the sharp one, because the loss isn't uniform: `finish_reason` rides a *separate* frame that parses fine, so it survives. Reproduced against the dev stack before the fix:

```
UPSTREAM SENDS:
  data: {...function:{"name":"get_weather","arguments":{"city":"Prague"}}}
  data: {...finish_reason:"tool_calls"}

CLIENT RECEIVES:
  data: {...finish_reason:"tool_calls"}
  data: [DONE]
```

A completion announcing a tool call with no tool call in it. Ollama's bare `{"error":"model not found"}` failed the same way, so the client saw nothing while the request log recorded a clean completion.

## The fix

The output- and error-bearing fields become `json.RawMessage`, read through `deltaText` / `argumentsText` / `chunkErrorMessage`.

That is the whole change. These frames now take the **ordinary path**, so masking, `strip_reasoning`, delivery accounting, the circuit-breaker verdict and the retirement verdict apply to them with no special handling — correct by construction rather than re-derived.

`deltaHasContent` no longer judges only the plain-string form of `content`, which otherwise rewrote a content-as-parts delta into a keep-alive and lost the text on the strip path. `carriesErrorObject`'s judgement is factored into `carriesErrorMember` so the observers, which hold the extracted member rather than the whole body, share one definition of what counts as an error.

**The existing suite passes unchanged**, which is the point: for shapes that already parsed, nothing is different.

## Why not #806's approach

#806 forwarded these frames *around* the typed parser. That meant hand-deriving everything the typed path feeds, and two review rounds found a different one of those broken each time — metering over-charged 8×, a failing provider escaped its circuit-breaker charge, reasoning leaked past a per-key strip contract, and a served stream still read as inconclusive to retirement. Each fix was another hand-derivation of work the typed path already does exactly. Widening the struct removes the second path instead of trying to keep it in step.

It also reaches cases the bypass could not: output living in `choices[1..n]`, which `observeDataChunk` already loops over.

## Telling readable output from output we cannot size

Widening the fields is not enough on its own, and two review rounds found why.

**Every part must be understood, not merely one of them.** `deltaText` called a parts array readable as soon as one element carried `"text"`, so a mixed array — `{"type":"thinking"}` beside `{"type":"text"}`, which is how Anthropic-style content blocks arrive — counted as read, and the reasoning-strip transform forwarded the whole delta verbatim to a caller whose key forbids reasoning.

**Unreadable output is still output.** An unreadable member was treated as "the provider may have delivered nothing". The caller *received* those bytes, so the provider delivered — saying otherwise metered a served request at zero while the provider billed for it, and told the retirement verdict the model never answers. `approxOutputBytes` sizes it by the strings inside; over-counting slightly is the safe direction when the alternative is charging nothing for a served request.

**`deltaTextPtr` is exactly what the `*string` field was** — a plain JSON string or nil. It had been handing `normalizeReasoningChunk` the *joined text* of a parts array, which the transform patched back as a plain string, destroying every non-text part: an image part vanishing from the caller's answer.

**`promptTextBytes` keeps its own extraction.** Sharing `deltaText` was wrong: it deliberately *skips* parts it does not recognise, because a vision request's base64 media must not be sized, while `deltaText` has to report failure to understand. They look alike and answer different questions.

`"error": null` was triggering the key-shape mask — `Error` is `RawMessage` now, so a present-but-null member is non-nil, and a gateway that stamps it on every chunk had the redaction regex run over the model's answer. `"error": false` and `"error": 0`, which several gateways use to mean *no* error, no longer mark a delivering stream failed. And an error member with no `message` is **named** rather than rendered raw, on both the observer and the probe path, because that string is persisted to `request_logs.error_message` and providers echo the offending input inside content-filter errors.

## Verification

Live against a rebuilt dev stack, one fake upstream per shape:

| upstream shape | result |
|---|---|
| tool `arguments` as an object | tool call **and** its `finish_reason` both delivered |
| `content` as an array of parts | delivered |
| `reasoning_content` as parts | delivered |
| Ollama bare-string error | 502, `error_kind=provider_error` — recognised instead of silently swallowed |
| spec-correct string content | unchanged |
| malformed bytes, then a good frame | garbage dropped, the good frame delivered |

No circuit-breaker transitions for any of the delivering providers.

Metering exactness is covered by unit test rather than live: the estimate charges quota and TPM but is deliberately never written to the request-log row, and the dev stack's stdout logging was dropping the INFO line that carries it. `TestTolerantChunk_MetersTheTextNotTheEnvelope` pins that the same five characters of output cost the same whether sent as a string or as parts, and fails if the extraction is removed.

Mutation-tested throughout: narrowing `deltaText` back to strings fails 5 tests and making it report everything readable fails 7; `argumentsText` rejecting the object form fails its table; `chunkErrorMessage` rejecting bare strings fails 2 and rendering the raw member fails 4; the mask on a raw nil check leaks a mangled answer; `deltaHasContent` trusting unreadable content leaks the chain of thought.

A differential harness over 48 SSE bodies confirmed that every shape which already parsed — plain string, empty, null, reasoning-only, spec-form tool calls, usage chunks, finish-reason normalisation and suppression, the empty-content strip, multi-choice, and all five `strip_reasoning` variants — is byte-identical to master, metering included.

### DIFF
--- a/internal/proxy/helpers.go
+++ b/internal/proxy/helpers.go
@@ -495,19 +495,10 @@ func promptTextBytes(body []byte) int {
 	}
 	n := len(req.Tools)
 	for _, m := range req.Messages {
-		var text string
-		if json.Unmarshal(m.Content, &text) == nil {
-			n += len(text)
-			continue
-		}
-		var parts []struct {
-			Text string `json:"text"`
-		}
-		if json.Unmarshal(m.Content, &parts) == nil {
-			for _, p := range parts {
-				n += len(p.Text)
-			}
-		}
+		// Same string-or-content-parts extraction the stream observers do, and
+		// deliberately the same helper: two copies of this drifted once already.
+		text, _ := deltaText(m.Content)
+		n += len(text)
 	}
 	return n
 }
--- a/internal/proxy/helpers.go
+++ b/internal/proxy/helpers.go
@@ -495,10 +495,25 @@ func promptTextBytes(body []byte) int {
 	}
 	n := len(req.Tools)
 	for _, m := range req.Messages {
-		// Same string-or-content-parts extraction the stream observers do, and
-		// deliberately the same helper: two copies of this drifted once already.
-		text, _ := deltaText(m.Content)
-		n += len(text)
+		// Deliberately NOT deltaText, despite the shapes looking alike. The two
+		// answer different questions and must not be merged: this one SKIPS a
+		// part it does not recognise, because a vision request's base64 media
+		// must not be sized, while deltaText has to report that it failed to
+		// understand a part — the reasoning-strip contract depends on knowing
+		// when the gateway cannot see inside a delta. Sharing them broke this.
+		var text string
+		if json.Unmarshal(m.Content, &text) == nil {
+			n += len(text)
+			continue
+		}
+		var parts []struct {
+			Text string `json:"text"`
+		}
+		if json.Unmarshal(m.Content, &parts) == nil {
+			for _, p := range parts {
+				n += len(p.Text)
+			}
+		}
 	}
 	return n
 }
--- a/internal/proxy/probe_error_frame_test.go
+++ b/internal/proxy/probe_error_frame_test.go
@@ -441,7 +441,12 @@ func TestErrorEnvelopeMessage(t *testing.T) {
 		// the frame into a streamChunk, which fails outright on this shape, so a
 		// local Ollama box in a hedged group reproduced the whole incident.
 		{"ollama bare string", `{"error":"model not found"}`, "model not found", true},
-		{"object without a message", `{"error":{"code":500}}`, `{"code":500}`, true},
+		// Named, not rendered raw. This message is persisted to
+		// request_logs.error_message, and providers echo the offending input
+		// inside content-filter and moderation errors — rendering the member
+		// verbatim would put the caller's prompt in the log, against the rule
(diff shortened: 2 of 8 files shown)


hugalafutro:
Splitting this up. Three review rounds established that the content-as-parts half cannot be settled by a shape heuristic: reasoning arrives as {"type":"thinking"} (Anthropic), {"text":..,"thought":true} (Gemini) and {"type":"reasoning.text"} (OpenRouter), so any rule for "is this content understood" is an allowlist against an open set — and strip_reasoning is a hard per-key contract where a miss leaks a caller's chain of thought.

Replacing with three narrower PRs: tool-call arguments (no reasoning risk, exactly sizeable, and the case that actually loses a tool call while its finish_reason survives), then the error member, then content-as-parts with the strip question decided deliberately rather than guessed.

PR #828, opened 2026-08-30
Put the Front Desk TOTP and OIDC login ceremonies behind the per-IP limiter

hugalafutro:
Front Desk mounted its per-IP request limiter on `/api/pair`, `/api/auth/admin-exchange` and `/api/logout` only, leaving `POST /api/totp/login` and `GET /api/auth/oidc/{start,callback}` reachable at unlimited request rate. The gateway binary applies the limiter across its whole `/api` tree, so the two binaries disagreed about the same handlers.

Found by a Strix (GLM 5.2) run on 2026-08-30, then verified against the code before anything was written.

## Why it is a bug and not a scoping choice

- The comment above the limited group (`internal/frontdesk/server.go:500`) already says these routes "ride the same per-IP limiter as the login ceremonies". The routing did not do that.
- Both handlers were **already given** the limiter (`server.go:301,306`) and simply never applied it.
- The `WebAuthnHandler` does exactly this for its own login routes (`webauthn.go:125-129`), so the pattern was established and two of the three ceremonies missed it.

Neither handler's own defense closes the gap. TOTP's `loginThrottle` counts **failures**, so it engages only after `maxFailures` and never bounds the rate of the attempts that reach verification before it does. OIDC `Start` writes a `webauthn_sessions` row on every request that reaches a configured IdP (`oidc.go:235`), reclaimed only by the hourly sweep, so an unbounded request rate is an unbounded row count.

## What changed

The limiter now sits **inside** `TotpHandler.Register` and `OIDCHandler.Register`, wrapping the public ceremonies only. Attaching it at the mount point would fix today's route table and leave the next reshuffle free to drop it again, which is how this arose; attached to the route, the protection travels with it and holds in both binaries.

`/api/totp/status` and `/api/auth/oidc/status` stay outside the limiter. They are the login screen's polls, they build no provider and touch no network or database, and the second test pins that they are not throttled.

## The scanner's proposed fix was rejected

Strix recommended applying the limiter at the `/api` route level "matching the gateway". That would regress the dashboard: Front Desk runs **5 rps / burst 10** (`cmd/frontdesk/main.go:231-232`) against the gateway's **30 / 60**, so a tree-wide limiter puts every authenticated dashboard and Bellhop device poll on a 5 rps budget, and a cold dashboard load bursts past 10 on its own. A passing test suite cannot see that.

Known trade: on the gateway these routes now sit under two limiters (the `/api` one and the handler's), so a login request spends two tokens. That is already true of WebAuthn login today, it only affects login ceremonies, and at 30 rps it is not a practical bound.

## Tests

`TestPublicLoginRoutes_RideThePerIPLimiter` drives each ceremony past its burst on a real Front Desk router, with `/api/pair` as a control so a failure tells "the limiter died" apart from "the scope regressed". Before the fix it failed exactly as the finding predicts (`totp/login` 400, `oidc/start` 400, `oidc/callback` 302 past the burst) while the control correctly returned 429.

`TestLoginScreenPolls_AreNotRateLimited` pins the other half of the scope. Mutation-tested by pulling `/status` into the limited group: the test fails, so it is not vacuous.

Green: 896 tests in `adminauth` + `frontdesk`, 2482 in `api` + `cmd` (no gateway regression), `golangci-lint` 0 issues, size gate clean, diff coverage **100%** (11/11).

### DIFF
--- a/internal/adminauth/oidc.go
+++ b/internal/adminauth/oidc.go
@@ -167,9 +167,19 @@ type oidcLoginState struct {
 // it. Mount on the same unauthenticated group as the WebAuthn/TOTP login routes.
 func (h *OIDCHandler) Register(r chi.Router) {
 	r.Route("/auth/oidc", func(r chi.Router) {
+		// Status is the login screen's poll: it builds no provider and touches
+		// no network or database, so it stays outside the limiter and a cold
+		// dashboard load cannot spend the ceremonies' budget on it.
 		r.Get("/status", h.Status)
-		r.Get("/start", h.Start)
-		r.Get("/callback", h.Callback)
+		// Start and Callback carry the per-IP request limiter the way the other
+		// login ceremonies do. Start writes a login-state row on every request
+		// that reaches a configured IdP, and only an hourly sweep reclaims
+		// them, so an unbounded request rate is an unbounded row count.
+		r.Group(func(r chi.Router) {
+			r.Use(h.ipLimiter.Middleware)
+			r.Get("/start", h.Start)
+			r.Get("/callback", h.Callback)
+		})
 	})
 }
 
--- a/internal/adminauth/oidc.go
+++ b/internal/adminauth/oidc.go
@@ -167,9 +167,11 @@ type oidcLoginState struct {
 // it. Mount on the same unauthenticated group as the WebAuthn/TOTP login routes.
 func (h *OIDCHandler) Register(r chi.Router) {
 	r.Route("/auth/oidc", func(r chi.Router) {
-		// Status is the login screen's poll: it builds no provider and touches
-		// no network or database, so it stays outside the limiter and a cold
-		// dashboard load cannot spend the ceremonies' budget on it.
+		// Status is the login screen's poll. It reads four settings keys and
+		// nothing else: no provider build, no outbound IdP call, no state
+		// write, and the settings behind it are TTL-cached, so it stays outside
+		// the limiter and a cold dashboard load cannot spend the ceremonies'
+		// budget on it.
 		r.Get("/status", h.Status)
 		// Start and Callback carry the per-IP request limiter the way the other
 		// login ceremonies do. Start writes a login-state row on every request
--- a/internal/adminauth/totp.go
+++ b/internal/adminauth/totp.go
@@ -88,7 +88,15 @@ func NewTotpHandler(
 func (h *TotpHandler) Register(r chi.Router) {
 	r.Route("/totp", func(r chi.Router) {
 		r.Get("/status", h.Status)
-		r.Post("/login", h.Login)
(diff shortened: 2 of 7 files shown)


hugalafutro:
Independent review round (fresh-context agent, asked to argue against the design rather than confirm it). Three findings taken, two deferred with reasons.

**Taken**

1. **The comment justifying the `/api/auth/oidc/status` exemption was false.** It claimed the handler touches no database; `Status` reads four settings keys (`oidc.go:199-202`). That sentence was the written rule a later maintainer would apply to a new route, so it mattered beyond tidiness. Reworded to what is true: no provider build, no outbound call, no state write, TTL-cached settings behind it.

2. **Applying that rule honestly selects a fourth route the first commit left alone.** `GET /api/webauthn/available` is public like the two status polls, but unlike them it runs an uncached `ListCredentials` per request (`webauthn.go:88-99`) against the embedded SQLite store. It is now in the limited group. Worth naming plainly: the first commit reasoned carefully about login-screen polls by name and still missed one in the same route table, which is a fair point against per-route opt-in and for a fail-closed tree-wide limiter.

3. **Mounting the same limiter twice charged two tokens for one request.** The second charge lands on a bucket the first just emptied, so once `rate_limit_ip_rps` drops below about 5 the second charge's delay exceeds `max_wait` and *every* login attempt 429s: an operator with 2FA or SSO cannot log in at all. This was already reachable through passkey login before this branch; the branch would have extended it to the remaining two ways in. `IPLimiter.Middleware` now marks the request as charged, keyed by the limiter so two different limiters still charge independently.

**Deferred, deliberately**

- `GET /healthz` (two uncached queries) and `GET /traefik/config` (open by design when `FRONTDESK_TRAEFIK_TOKEN` is unset, and its `RecordConfigPoll` clears `staleNotified`, so an unauthenticated caller can keep the Traefik-stalled watchdog quiet) are both unlimited. Real, pre-existing, and each needs its own decision about breaking token-less deployments and health probes. Not smuggled into a login-hardening PR.
- The `s.ipLimiter != nil` guards at `frontdesk/server.go:503,514` are dead now that three `Register` methods dereference it unguarded. Cosmetic inconsistency, left alone.

**Calibration note:** the TOTP half of the original finding is defense-in-depth rather than a closed hole. `loginThrottle.Allowed` is checked before any work (`totp.go:345`), so a flooding IP was already refused after 5 failures with exponential backoff. The OIDC half is the one that was genuinely unbounded, because `Start` writes a login-state row reclaimed only by the hourly sweep.

Verification after the round: 739 tests in `ratelimit` + `frontdesk`, 4547 in `api` + `proxy` + `cmd`, `golangci-lint` 0 issues, size gate clean, diff coverage **100% (21/21)**. Both new production behaviours mutation-tested: removing the `/available` limiter fails the route test, removing the charge-once guard fails the limiter test.

PR #829, opened 2026-08-30
Report a model whose circuit keeps reopening

hugalafutro:
The queued follow-up from #827 (design point 10): a model circuit that keeps reopening should escalate instead of failing silently forever.

Since #827 keyed circuits per model, "this one model keeps breaking" is a fact the gateway holds and never acts on. A model whose circuit opens, cools down, half-opens and opens again costs every group that names it a doomed attempt plus a cooldown, on repeat, until an operator happens to notice.

## What it does

The breaker counts transitions into Open per circuit inside a 24h window and reports once when a third lands: a `circuit_breaker.unstable` event and a warning log, both naming the model.

- An open past the window starts a new count rather than extending the old one, so a report is always drawn from one run of recent failures rather than unrelated outages months apart. Same shape as `goneStreak.strike`, for the same reason.
- Crossing the threshold clears the window, so one window reports once. A model that stays broken reports again on its next three opens, not on every open.
- Counters live on the existing circuit struct, so they inherit #827's ~256-closed-per-provider eviction bound. No new map, no new eviction rule, no persistence.

## It reports, and never retires. That is the design, not a shortcut

Retiring a model asserts the provider no longer serves it, and only a request **refused by name** establishes that. A model failing that way is already nominated by the gone-classified streak on its first refusal (#596/#598). What is left for this event is the model that fails some *other* way, repeatedly: timeouts, 5xx, a provider having a bad week. Disabling that model would take a working one out of routing on evidence that says only "upstream is unwell".

**The plan I wrote for this follow-up said to hand the model to the retirement probe and let the probe adjudicate. That turned out to be unimplementable, and the reason is worth recording:** `model_gone.go` refuses to probe a model whose circuit is open, and an open circuit is by definition this model's state at the instant the escalation fires. Every nomination from here would hit that gate, be postponed, and be dropped. `probeModel` checks the same thing independently, so there is no entry point that avoids it. Firing later, on the next half-open, was considered and rejected: at half-open the breaker is already sending a real trial request, and if that request were refused by name the gone-streak would nominate it anyway. There is no population of models for which "reopened three times" is better evidence of retirement than what already runs.

## Alerting

Catalogue entry is **default-off**, matching the Discovery backlog events and for the same reason: an instance upgrading with a provider that has been flaky for weeks would otherwise be alerted on the third open about a condition its operator already lives with. No migration, so existing installs are untouched; operators opt in under Settings, Alerts.

## Test-guard change worth a second look

`TestCatalogTypesAreEmitted` asserted that every `circuit_breaker.*` catalogue type is emitted, detecting them by the state string passed to `publishEvent`, since those types are built by concatenation. This event names no circuit state, so it is published as a literal. The guard now accepts either form. The invariant is unchanged and the literal check is the same evidence it already requires of every other event type; verified by mutating the emitted type, which still fails the guard.

## Verification

TDD throughout. Window arithmetic is unit-tested directly on `noteOpen`, because the escalation it governs is a once-a-day event and driving 24 hours of transitions through the breaker would need a clock seam in production code to observe. One integration test drives the real breaker through three open transitions and asserts the event, its severity and the log line.

Mutations, all killed: dropping the once-per-window reset (fourth and fifth opens re-report), making the window never expire (opens a day apart add up), removing the escalation call entirely (no event), and renaming the emitted type (catalogue guard fires).

335 tests green in `alert` + `failover`, 2655 across `alert` + `failover` + `api`, `golangci-lint` 0 issues, size gate clean, diff coverage **100% (35/35)**. Wiki updated in both places that carry the event catalogue.


hugalafutro:
Independent review round (fresh-context agent, asked to argue the design was too weak). It found a real bug in the first commit, and I took six of its eight findings.

**The serious one: the reporting cadence was two orders of magnitude off what every piece of copy around it claimed.**

`noteOpen` cleared the window when it reported, which starts a fresh window on the very next open. A model failing every cooldown reopens about once a minute, so it reached three opens again within minutes and reported every few minutes, each time saying it had "broken 3 times in 24h" — a sentence its own cadence disproves. Roughly 480 reports a day for exactly the population this feature targets. The window now keeps running and is marked reported, so a report costs a full 24 hours before another can follow. Pinned by a test that drives 40 opens through one window and a mutation that restores the old behaviour.

**Also taken**

- **The event named a model but debounced on its provider.** It carried `provider_id` as its only debounce identity, and the dispatcher probes most-specific-first with `provider_id` ahead of `model_id`. Two models on one provider cross their thresholds together (they fail for the same upstream reason), so the second alert was dropped and the model behind it could stay unreported indefinitely — the exact failure per-entity debouncing exists to prevent. The event now carries `model_id`, and `model_id` leads the probe order. Blast radius checked: `failover.sync_error` carries `model_id` and no provider id, and `circuit_breaker.open`/`closed` carry no `model_id`, so nothing else changes granularity.
- **No i18n label.** The `settings.alerts.event` block is strictly one key per catalogue entry, so the Alerts page would have shown a checkbox labelled literally `circuit_breaker.unstable`, untranslated in every locale. Added and hand-translated into all 29 dashboard locales, each reusing the circuit-breaker noun that locale already chose.
- **The window was written three ways** — `24h0m0s` in the metadata, `24h` hardcoded in the message, `24h` documented in the wiki. All three now derive from the constant.
- **The message overclaimed.** "Keeps returning to service unhealthy" is not what the counter establishes: a success does not reset it, so three recovered blips across a working day land there too. It now says only what was counted.
- **The default-off rationale was wrong about its own surface.** `DefaultOn: false` gates outbound notification; the dashboard's event feed is not catalogue-gated. Corrected in the catalogue comment and the wiki.
- **The integration test could not fail the way that mattered** — it asserted an event arrived, not that none arrived before the third open, so reporting unconditionally would have passed. It now steps one open at a time and pins the `opens`, `window` and `model_id` metadata the wiki documents as contract.
- Stale header comment on the relaxed catalogue guard, fixed.

**On the reviewer's main argument, which I partly accept.** It agreed with "never retire" and verified the probe-gate claim independently, but argued the PR opens by framing the cost as wasted requests and then does not remove one: with a 60s cooldown a broken model still burns ~1440 real user requests a day as failed half-open probes. That is fair. The fix it proposes — exponential cooldown backoff for a circuit whose open counter is climbing — is genuinely the behaviour that saves those requests, and the counter this PR adds is exactly the state it would consume. I have not folded it in, because `cooldownOverride` currently means "quota pin" and is gated on `circuit_breaker_quota_pin_enabled`, so a backoff needs its own field and a precedence rule against a pin. That is a design decision, not a mechanical addition. **Queued as a follow-up**, and I have reworded nothing in the PR body to hide it: visibility is the gain here, and the request waste is still open.

Verification after the round: 336 tests in `alert` + `failover`, 4547 in `api` + `proxy` + `cmd`, 6016 frontend tests, `tsc -b` clean, `make i18n-check` green across all targets, lint 0 issues, size gate clean, diff coverage **100% (46/46)**. Three mutations killed: restoring the window-clearing cadence, reporting on every open, and reverting the debounce order.

PR #831, opened 2026-08-31
feat(failover): probe backoff for chronically failing circuits; per-model debounce for open/closed alerts

hugalafutro:
Two of the three items queued after #827 and #829. The third (negative caching in settings) is a different package and follows separately; it also became more pressing here, see "cost" below.

## 1. A circuit whose probes keep failing backs off

#829 made a chronically-reopening circuit report itself, and removed none of the waste it reports. The half-open probe is a live user request (`IsOpen` hands it to the next request for that model), so a model that is broken all day wastes one real request per `circuit_breaker_cooldown`: about 1,440 a day at the 60s default.

Now every probe that fails doubles the cooldown before the next one, capped by a new `circuit_breaker_backoff_max` (default **15 minutes**), and a probe that succeeds closes the circuit and resets the count. At the defaults the waits run 1, 2, 4, 8 minutes and then hold at 15, so the same broken model wastes under a hundred requests a day, and a model fixed upstream is back within a quarter of an hour with nobody touching anything.

**Why 15 minutes and not longer.** The first cut shipped a 1h ceiling and the review argued it down, correctly: a probe is the only way a model with *no* healthy sibling ever gets tried again, so the ceiling is also the worst-case time such a model stays unreachable after it has actually recovered, and the provider-wide verdict holds for as long as its blocking circuits do, so it bounds how long a partially broken provider's healthy models stay skipped. 15 minutes keeps ~93% of the saving, bounds both of those to a quarter of an hour, and happens to be the failover page's fuse-animation window, so a backed-off fuse still visibly burns. Raise it if you can wait longer.

**What counts is a failed probe, not an open.** #829's opens-in-24h counter also counts a circuit that legitimately closed on a good probe and was later charged back to threshold by fresh failures. Doubling on that would slow the recovery of a model that had two real, separate outages in one day, which is exactly the model an operator wants back promptly. A failed probe is the breaker's own experiment failing; it is direct evidence the model is still broken.

**The longest cooldown in force governs.** A circuit can carry a quota pin and a backoff at once; it serves whichever reaches further, and `quota_pinned` / `backed_off` each say whether their override is in force. A pin is floored at the cooldown in force when it is stamped, *after* the pin ceiling has been applied (the same order `ApplyQuotaPins` already used), so neither a short quota reading nor a low `quota_pin_max` can pull a backoff in; releasing a pin falls back to the backoff, not the base. The stamped backoff never serves less than the configured cooldown, so raising `circuit_breaker_cooldown` past it makes the setting govern and drops the flag. Each of those sentences is one test; the first cut got three of them wrong and the review caught all three.

**Cost.** The backoff is stamped at the open transition and stored, like the pin, so per-provider walks read nothing per circuit. The two kill switches are read once per walk, lazily, only when a circuit actually carries an override or a backoff (`cooldownReads`): the healthy path still costs the one settings read it always did, and a provider with a hundred backed-off circuits costs one more, not a hundred. That matters because a settings key nobody has set is an uncached SELECT on every read (the queued negative-caching item), and these walks run under the breaker lock on the request path and the scrape path. The read-cost guard now counts the switch reads too; the first cut's per-circuit read passed the old guard, which only counted the cooldown.

**Surfaces.** Two settings (both allowlists), `backed_off` + `failed_probes` on the status API, on the `circuit_breaker.open`/`closed` events and on the open log line; `cooldown_ms` and `next_retry_at` on every backed-off or pinned open event, and the event message says so (`... (backing off after 3 failed retries, next retry in 8m)`), since the message is all an outbound alert renders; a toggle and a 1..240-minute slider in Settings mirroring the quota-pin pair; a tooltip on a backed-off failover entry (ranked below a provider-wide skip and a quota pin, which both say more); 29 locales; the wiki, both Configuration pages and the API reference, including a new note that the provider verdict now persists as long as its blocking circuits' backoffs. The tooltip carries no count, so it needs no plural forms.

## 2. `circuit_breaker.open` / `closed` debounce per model, until the provider is skipped

Since #827 those events are per model circuit, but they carried only `provider_id`, so the alert dispatcher debounced them per provider and the second model failing on one provider inside the 5-minute window was silently dropped. They now carry `model_id` (which leads the dispatcher's debounce keys since #829) **while the provider is still serving**. Once the verdict says the provider itself is skipped, `model_id` is omitted and the event debounces on `provider_id` again: the verdict lapses every time a blocking circuit's cooldown elapses, which lets one more sibling through to open, so a fifty-model provider outage keyed per model would have notified fifty times for one fact. The review caught that the first cut's "at most `span` alerts" claim was false.

## Not changed, and why

- **The nav badge** keeps its three tooltip lines. With a 15-minute ceiling a backed-off provider is "back shortly" in the sense the amber line already means; a fourth line would cost a pluralised key across 29 locales for a distinction the entry tooltip and the event already make.
- **The Prometheus gauge** still buckets by state and cannot tell a 60s circuit from a 15m one. A cooldown gauge needs a per-model label decision the gauge has so far avoided; queued.
- **Half-open concurrency** is unchanged: many concurrent requests can be handed the same probe, and a success arriving after a failure has re-opened the circuit is discarded. Pre-existing, and with HalfOpenMaxProbes=1 the chance that a 90%-healthy model ratchets past n failed probes is 0.1^n, so it is not the "ratchets to the ceiling" the review feared, but it is real for a coin-flip model.

## Housekeeping

`circuitbreaker.go` crossed the 800-line ceiling, so the quota-pin block moved to `circuit_quota.go`. Not verbatim: `applyQuotaPin`'s clamp order and the floors are the fixes above.

## Tests

`internal/failover/circuit_backoff_test.go`, one test per behaviour and one per review-caught bug: doubling with the enforced cooldown checked through `IsOpen`, the default and a configured ceiling, a ceiling below the base adding nothing, a raised base governing and dropping the flag, reset on a successful probe, the kill switch releasing and restoring, the pin floor on both stamp paths, the pin ceiling applied before the floor, a pin stamped under a disabled backoff never outranking it once re-enabled, a released pin falling back to the backoff (log line included), the switch reads not scaling with backed-off circuits (Status and a sibling `IsOpen`), and the event: `model_id` present, absent once the provider is skipped, `cooldown_ms`, `next_retry_at` only under an override, and the message suffix. Every fix mutation-tested by hand against its named test.

Live on the dev stack (comment below, first cut): 30s → 60s → 120s → 120s at a 2m ceiling, on both circuits of a two-member group of unreachable hosts, with the log lines and the kill switch verified past the status endpoint's 5s cache. Re-run on the reviewed build in a second comment.

Frontend: 478 tests across the Settings page and failover entries. `tsc -b`, biome, `make i18n-check`, size gate and lint all clean.


hugalafutro:
## Live on the dev stack (this commit, `0a26c32a`)

Settings for the run: threshold 1, cooldown 30s, backoff ceiling 2m. A two-member `hotel/backoff-test` group whose members are both unreachable hosts (connection refused), so every request charges both circuits. One request per cooldown, timed from the status API's `next_retry_at`:

| request | result | `cooldown_ms` | `backed_off` | `failed_probes` |
|---|---|---|---|---|
| 1 (closed→open) | 502 | 30000 | absent | absent |
| 2 (probe failed) | 502 | 60000 | true | 1 |
| 3 (probe failed) | 502 | 120000 | true | 2 |
| 4 (probe failed) | 502 | 120000 (ceiling) | true | 3 |

Identical on both circuits. The `half-open→open (probe failed)` log lines carry the same three fields (`cooldown_ms=60000 backed_off=true failed_probes=1`, then 120000/2, 120000/3), and the `closed→open` line reports `backed_off=false failed_probes=0`. #829's `model keeps reopening its circuit` warning fired on the third open of each, as it should, since the two features count different things.

Kill switch, read after the status endpoint's 5s cache: `circuit_breaker_backoff_enabled=false` took a circuit at 60000/`backed_off` to 30000 with no `backed_off` and `failed_probes` still 1; flipping it back restored 60000/`backed_off`. Both providers then reset to closed via the per-provider reset, and the dev settings, group and key are removed.

hugalafutro:
## Live on the dev stack, reviewed build (`1fbec61f`)

Same setup as the first run (threshold 1, cooldown 30s, ceiling 2m, two unreachable members), plus the review's raised-base case.

| step | `cooldown_ms` | `backed_off` | `failed_probes` |
|---|---|---|---|
| open | 30000 | absent | absent |
| probe failed | 60000 | true | 1 |
| probe failed | 120000 | true | 2 |
| `circuit_breaker_cooldown` raised to 5m | 300000 | absent | 2 |
| back to 30s | 120000 | true | 2 |
| `circuit_breaker_backoff_enabled=false` | 30000 | absent | 2 |
| back to true | 120000 | true | 2 |

Every reading taken after the status endpoint's 5s cache. The raised base governing with the flag dropped (and the count kept) is review finding B3 fixed, observed rather than unit-tested only. Log lines carry `cooldown_ms=60000 backed_off=true failed_probes=1`, then 120000/2, and the first open `backed_off=false failed_probes=0`; #829's "keeps reopening" fired on the third open as before. Circuits reset, settings/group/key removed; the dev stack is back to its prior three breaker keys.

hugalafutro:
## Review rounds

Two independent adversarial reviews of the first cut (`0a26c32a`), one on the Go core and one on operator-facing behaviour, then a scoped re-review of the fix commit. Dispositions:

**Fixed in `1fbec61f`**
- B1/B3/B4 (pin vs backoff precedence: a pin stamped under a disabled backoff outranked it once re-enabled; a stamped backoff kept serving below a raised base while reporting `backed_off`; the retarget floor lost its base guard). The longest cooldown in force governs now, and a backoff only counts beyond the base.
- B2: `applyQuotaPin` clamped to the pin ceiling after the floor, so a low `quota_pin_max` could pin below a backed-off cooldown. Ceiling first, as `ApplyQuotaPins` already did.
- P1/P2: the kill switch was a settings read per backed-off circuit under the lock, and the read-cost guard could not see it. `cooldownReads` hoists both switches per walk, lazily; the guard counts them.
- D2: "at most `span` open alerts per outage" was false; `model_id` is dropped from an `open` once the provider is skipped.
- UX 1 (1h default ceiling regressed recovery latency for a sibling-less model): 15m. UX 2 (verdict persistence undocumented): documented. UX 4 (backed-off event carried no deadline): `cooldown_ms` + `next_retry_at`, and the message says so. UX 5/8/9/10/11/12 (docs and copy): fixed, 29 locales retranslated. UX 7 (1440-position slider): 240. UX 13: disabled-controls test added.

**Fixed in `3c7c26e8`** (from the re-review)
- The backoff suffix was added to a pinned-and-backed-off open even when the pin was the longer wait, blaming the model for the provider's spent window. Only when the backoff is the value in force now.
- The `model_id` drop applied to `closed` events too, collapsing two recoveries on a still-skipped provider into one alert. A close always carries it.
- `shortDuration` rendered 1h30m as `90m` and sub-second cooldowns as `0s`. Comments still promising the number and the flag "can never disagree" now say which flag says what. Tests for the retarget floor against a raised base (B4 proper), the suffix rule, the close on a skipped provider, exact switch-read counts, and the formatter.

**Deliberately not changed**
- Nav badge gets no backoff line (a pluralised key across 29 locales for a distinction the entry tooltip and the event already make, and at 15m "back shortly" holds).
- No Prometheus cooldown gauge (per-model label decision; queued).
- Half-open concurrency (pre-existing; a 90%-healthy model ratchets past n failed probes with probability 0.1^n).
- No decay on `failed_probes` beyond the reset on close, which is the only recovery that matters.
- `omitempty` on `backed_off`/`failed_probes`, matching `quota_pinned`.
- Direct (non-hotel) requests can count a failed probe: the request did fail against a model the breaker was already doubting.

Both live runs on the dev stack are in the comments above; the second is on `1fbec61f`, and `3c7c26e8` changes only event text, the `closed` key rule, comments and tests.

PR #832, opened 2026-08-31
perf(settings): cache an absent key like a value

hugalafutro:
The third item queued after #827: the settings cache only ever cached rows, so a key nobody has set was an uncached `SELECT` on every read for the life of the process. That is the common case, not the edge: nearly every runtime setting is read on the proxy's per-request path with a default the operator never overrode, and #831 made it concrete by putting two more such reads under the breaker lock.

## What changes

`GetWithDefault` and `GetChecked` now cache an absence (`cacheEntry.missing`) for the same TTL as a value, under the same per-key generation guard, and every write path evicts it: `Set`, `SetMany` and `DeleteKey` evict on both sides of the write exactly as they do for a value, and `SetTx` / `DeleteKeysTx` rely, as they already did, on their callers' `InvalidateCache` / `NotifyDeleted` after commit (both callers do; that contract is now written on `SetTx`, and tested over a cached absence). So the first write after a miss is seen at once. A real database error is still never cached, and still warns. `IsCached` reports a cached absence as cached, which is what it is; its one caller outside the package is the proxy's `cache_hits.settings` request-log flag, which therefore turns true for a default read instead of being permanently false, which is the optimisation showing.

One consumer moved as a result: the config-sync import invalidated synced settings only at the very end of `postImportRefresh`, behind a discovery pass that can run for minutes. With absences cached, a key the primary set for the first time would have stayed "unset" on the member until that pass ended (or the TTL ran out), so the invalidation now runs right after the provider/model caches are dropped, before discovery, and a removed key gets `NotifyDeleted` alone rather than an `InvalidateCache` that re-read a row that no longer exists followed by a second notification.

The generation guard matters more for an absence than for a value. A reader that observed "no row" while a `Set` was landing must not install "absent" over the write, or the setting the operator just saved is invisible for a full TTL; `install` applies the guard to both kinds of entry through one path.

## Not changed

`WarmCache` still installs only the rows that exist. Absences fill in lazily on first read, one `SELECT` per key per TTL, which is the cost this PR reduces the per-read cost to; pre-seeding every allowlisted key at startup would save that first read at the price of a second list to keep in sync. The 30s TTL and the change-notification path are untouched, so the wiki's "changes take effect immediately" stays true.

## Tests

`internal/settings/settings_negative_test.go`: an absence is cached (a row inserted behind the repository's back stays invisible for the TTL, exactly as a changed value would); both readers share it; `Set`, `SetMany` and `DeleteKey` see through it at once, and the absence a delete leaves behind is cached in its turn; it expires with the TTL; and the generation guard blocks a late absence, made deterministic the way the existing guard test is (the reader's `SELECT` parked behind `LOCK TABLE ... ACCESS EXCLUSIVE`, an eviction landed, the lock released). Each of the four load-bearing lines mutation-tested against its test. The settings-consuming packages (api, proxy, quota, frontdesk, ratelimit, cmd) pass unchanged.


hugalafutro:
## Live on the dev stack (`899ac770`)

The breaker reads `circuit_breaker_threshold` on every failure, so with the key unset it is the first thing this change caches as absent. Observed through a two-member `hotel/` group of unreachable hosts:

| step | result |
|---|---|
| key unset, one failed request | circuit `closed`, `consecutive_fails=1`: the default of 5 applied, and the absence is now cached |
| `PUT circuit_breaker_threshold=1`, one failed request | circuit `open`: the write evicted the cached absence and was honoured on the very next request |
| `DELETE` (reset to default), one failed request | circuit `closed`, `consecutive_fails=1`: the reset evicted the cached value and the default applied at once |

Circuits reset between steps; group and key removed afterwards; the dev settings are back to the three breaker keys that were there before.

hugalafutro:
## Review round

One adversarial review (fresh context, told to argue against the change). Its bottom line: the mechanism is sound, no interleaving produces an unbounded stale absence, because an absence can only be installed by a `SELECT` that returned no row, and everything that creates a row through the repository evicts. Twelve smaller items, dispositions:

**Fixed in `24056037`**
- Config-sync import invalidated synced settings only after a discovery pass that can run for minutes, so with absences cached a key the primary set for the first time stayed "unset" on the member until then. Invalidation now runs with the other cache drops, before discovery; a removed key gets `NotifyDeleted` alone instead of an `InvalidateCache` (a wasted re-read of a deleted row) plus a second notification.
- `cacheEntry`'s comment and the test comment claimed `Set`/`SetMany`/`DeleteKey` were the write paths; `SetTx` and `DeleteKeysTx` are how every operator save and every sync lands, evicting via their callers after commit. Both now say so, `SetTx` carries the evict-after-commit contract in its doc (as `DeleteKeysTx` already did), and `SetTx` + `InvalidateCache` over a cached absence has its own test.
- The two raw-SQL settings writers in config sync (`_fleet_last_source_gen`, `_fleet_unapplied_*`) now say the repository cache is not evicted for their keys, which is fine only while they are read raw, as they are.
- TTL test margin 100ms → 250ms.

**Noted, not changed**
- `IsCached` is true for an absence; its one outside caller is the proxy's `cache_hits.settings` request-log flag, which now reads true for a default read rather than being permanently false. That is the optimisation showing; the PR body says so.
- `WarmCache` writes the cache outside `install`; a reader could only install an absence over a warmed value if a row appeared without a generation bump, which needs raw SQL before startup completes. No live path.
- AGENTS.md's package-table note is a local, gitignored file; it is updated on this machine and not part of the diff.
- One assertion in the guard test is weak (`v != "default"`); the load-bearing one is `!IsCached`.

Live check on the dev stack in the comment above; api (2,320) and settings (62) suites pass, lint clean.

PR #833, opened 2026-08-31
feat(bellhop): notify for every Front Desk alert switched on under Alerts (0.9.13)

hugalafutro:
Bellhop: the alert toggles now decide which Front Desk alerts reach the phone.

## What was wrong

Switching "Member drained or activated" on in Bellhop's Alerts screen and rebuilding the fleet produced no notification, and it was not the phone killing the app. Traced hop by hop during the 2026-08-31 rebuild: Front Desk emitted `member.state_changed` for all four members, the Apprise dispatcher sent them, and the self-hosted ntfy server holds all of them in its cache for Bellhop's UnifiedPush topic. The last hop discarded them by design: `BellhopPushService.onMessage` read the payload only for the test marker, then ran the fleet poll, and `diffFleet` notifies only on a real UP→DOWN or DOWN→UP health edge ("Drain/activate is an operator action (not an alert)"). So the toggle configured Front Desk's outbound alerts, Front Desk honoured it, and Bellhop threw the result away. "FD → ntfy.sh" worked because the ntfy app displays the payload; Bellhop never did.

## What changes

The push stays a wake, not a data source (the ntfy distributor hands over the message body alone, with no event type or severity, and PR #422's reasons for not trusting a payload still hold). What the wake poll does is widened: after the fleet, it reads Front Desk's alert picker (`GET /api/alert/selection`, readable by any paired device) and one page of the event log (`GET /api/events`, newest first), and every event after the last one seen whose type the picker has on becomes a notification carrying Front Desk's own sentence, titled with the same catalogue name the Alerts screen shows. The periodic Layer-2 poll does the same, so a phone without push still gets them within a poll.

- `diffEvents` (pure, beside `diffFleet`): silent on the first poll (records the newest event, replays nothing), orders by `created_at` as an instant rather than text (Front Desk writes RFC3339 at whatever fractional precision the row has, and text order lies at whole seconds), lets the id decide same-instant ties the way Front Desk's own `ORDER BY created_at, id` does, caps at the newest five per poll so a phone that was off for a day is not paged once per line, never rests the cursor on a row whose time cannot be read, and advances the cursor past switched-off events too, so switching a type on later does not replay them.
- `EventCursor` persists in `MonitorStore` under the same active + session-epoch gate as the health snapshot, and `clear()` wipes it with the rest.
- Bellhop's own `WentDown` / `Recovered` / `WentStale` stay on regardless of the picker (they are the Layer-2 promise that needs no Front Desk alert configuration). Where Front Desk's own event for the same member's outage or the same drift is being posted in this very poll, Bellhop's row is not, so one outage is one notification. The yield is exact, never inferred: a failed picker or log read, a picker with the type off, a first poll still recording its baseline, or a Front Desk event pushed past the per-poll cap all leave Bellhop's own alert standing. A failed log read also keeps the cursor where it was.
- The log is read with `since=<cursor>` (Front Desk applies `created_at >= since`), so a page is normally a handful of rows; a first poll asks for `limit=1`. At most five Front Desk events are posted per poll, oldest first so the newest is the top row; older ones past the cap are history and stay in the app's event log.
- On app start, if a layer is on and no cursor exists yet (a fresh enable, or the upgrade from 0.9.12), Bellhop runs one immediate poll to record the baseline, so the first push after that is diffed instead of spent recording it.
- `FleetNotifier` renders a `FrontDeskEvent` on one of two new channels: "Front Desk errors" (high importance, pages) for error severity, "Front Desk alerts" (default importance) for everything else, so muting Front Desk's events never touches Bellhop's own health rows and vice versa. Rows are tagged by type, member and severity: a repeated warning about the same member updates in place, while a drain (warning) and the re-activation (info) are two rows, because Android keeps a row's original channel on update and the operator should see the drain. `eventTypeLabelRes` is the non-Compose half of `eventTypeLabel`, so the title is the label the Alerts screen uses; an unknown type falls back to its raw code, as the screens do.

Cost: two more requests per background poll (picker + one events page), to the same host on the same connection; the poll tests' request-count assertions record the change (3 → 5, 4 → 6) and say why. The push-test notification body and the wiki's push section no longer claim Bellhop only alerts on health and drift; ten locales retranslated.

## Not changed

No persistent-notification / foreground-service option. The evidence says the messages reached the phone's ntfy app and were dropped by Bellhop's own rule, not by OxygenOS, so that option would not have produced this alert. If the OS-kill theory still needs testing after this ships, leave the phone idle for an hour or two without opening Bellhop and press Front Desk's **Send test**: that notification is rendered from the payload and only arrives if the app is alive.

## Tests

`FleetEventDiffTest` (the pure rule: first-poll silence, cursor advance, picker filter, instant-not-text ordering, the id deciding same-instant ties the way Front Desk orders, oldest-first output, the cap, an empty page, an unreadable row never becoming the cursor), `MonitorStoreEventCursorTest` (round trip, the active gate, the stale-session gate, clear), `FleetEventPollTest` against MockWebServer (a switched-on event comes back as Front Desk's sentence and advances the cursor; a switched-off one advances the cursor only; Front Desk's `health.down` replaces Bellhop's own edge for that member and `config.autosync_stale` replaces the drift edge; the local edge stands for another member's event, when the picker has health off, on the upgrade poll with a snapshot but no cursor, when the event is pushed past the cap, when the picker read fails, and when the log read fails; `since`/`limit=1` on the wire; a dead token never reaches the log), `FleetNotifierEventTest` under Robolectric (title, body, the errors channel for errors only, drain and re-activation as two rows while a repeated drain is one, fleet-wide kinds keeping their rows, raw-code fallback). Existing poll and backstop fixtures gained the two responses.

`compileDebugKotlin`, `testDebugUnitTest`, `ktlintCheck`, `lintDebug` green; `make i18n-check` clean for the ten Android locales.


hugalafutro:
## Review round

One adversarial review of the first cut (`2e503f2d`), fresh context, told to argue against it. Fourteen findings; dispositions in `be499a87`:

**Fixed**
1. Upgrade poll (snapshot kept, cursor new) suppressed Bellhop's own `WentDown` while posting nothing. The yield is now exact: only a Front Desk event for the same member (or the same drift) actually being posted in this poll silences the local row. A first poll, a failed picker or log read, a picker with the type off, or an event pushed past the cap all leave Bellhop's own alert standing. Tested for each.
2. `config.autosync_stale` double-notified beside Bellhop's own `WentStale`; the same exact yield covers it.
3. The per-poll cap could drop the very event the local edge had yielded to; with the exact match that cannot happen (the event beyond the cap is not posted, so the local edge is). The log is also read with `since=<cursor>` (`limit=1` on a first poll) instead of fifty unfiltered rows, which answers #10.
4. Alerts are posted oldest first, so the newest is the top row and, for rows sharing a tag, the survivor.
5. Same-instant tie-break follows Front Desk's id order (`id > cursor.id`) instead of inequality, which re-reported a lower id every poll.
6/7/8. Front Desk events get two channels of their own: "Front Desk errors" (high) and "Front Desk alerts" (default, everything else), so muting them never touches Bellhop's health rows and vice versa; tags include severity, so a drain (warning) and the re-activation (info) are two rows rather than one row stuck on the warning channel.
11. On app start, if a layer is on and no cursor exists, Bellhop runs one immediate poll to record the baseline, so the first push after an enable or this upgrade is diffed instead of spent recording it.
12. The cursor never rests on a row whose time cannot be read.
13. Tests added for the upgrade case, the other-member case, the drift overlap, the cap-vs-yield case, the reverse-id tie, the errors channel, severity tagging, fleet-wide tags, the wire query, and the seed decision.
14. Docs: the wiki, the PR body, both KDocs and the push-test body no longer claim the picker decides everything; Bellhop's own down/recovered/drift rows stay on regardless, and the text says so.

**Deliberately not changed**
9. Bellhop's own health and drift notifications stay on even when the matching Front Desk type is off. They are the Layer-2 promise that needs no Front Desk alert configuration at all, and the docs now say that plainly instead of implying the toggles govern them.

hugalafutro:
## Re-review of `be499a87`

Scoped re-review (fresh context): 10 of the 14 findings fixed outright, 3 partial with the remainder stated in code, 1 by design with the docs corrected; 12 tests kill a revert; 17 targeted checks correct, including the two that decide whether the health dedupe works at all: Front Desk stamps `member_id` on `health.down`/`health.up` with the same id `/api/members` renders, and the event row is inserted before the push is published, so a push-woken poll always sees both the DOWN badge and the event together.

New items, all low, taken in `5a9cd247`: the baseline seed is gated on notification permission (otherwise it re-enqueued on every foreground while denied); the events-request test asserts the actual `since` value and its takes are bounded so a short poll fails instead of hanging; the severity-in-tag rule documents the fleet-wide types whose severity tracks state; the seed's coalescing with a concurrent push wake is stated in its KDoc.

Left as documented: a periodic (not push-woken) poll that lands in the sub-millisecond window between Front Desk marking a member unhealthy and inserting the event can post Bellhop's own row and then Front Desk's on the next poll, two rows instead of one; before this PR that window swallowed the outage, so the trade is in the right direction. A burst of more than fifty events since the cursor posts the newest five of the newest fifty and moves on.

PR #969, opened 2026-09-09
chore(release): 0.9.104 and Bellhop 0.9.15

hugalafutro:
## What

Patch release of everything that publishes from this repo.

| file | from | to | releases |
|---|---|---|---|
| `.version` | 0.9.103 | 0.9.104 | `model-hotel` and `model-hotel-frontdesk` images |
| `android/.version` | 0.9.14 | 0.9.15 | Bellhop APK |

## Why

The published 0.9.103 images were built on 4 September against grpc v1.83.1, which carries CVE-2026-84445. Master has been on v1.83.2 since `b35ef87c`, and the advisory names that as a fixed version, so the fix is merged and only needs publishing. This release is what puts it in the images, which clears both open code-scanning alerts and the scanner issue #967 behind them.

It also ships everything merged since 0.9.103, most recently the row stepper in the detail modals (#968).

Bellhop has no app changes in this bump. It rides its own version file and gets the matching patch so the two trains stay in step.

## Verification

- Fleet rebuilt to the current master before this branch was cut: all four members and Front Desk healthy, Docker stats intact on each, Front Desk marking every member healthy.
- Live model sweep against the rebuilt fleet: 304 pass, 8 weak, 24 fail, nothing regressed against the previous run, members agree on every probe. The failures are the known drained NeuralWatt fixture.


hugalafutro:
Closing unmerged: the release is being held for now. The version bump lives on chore/release-0-9-104 if it is wanted later.

hugalafutro:
Reopening: master already carries grpc v1.83.2, which is the fixed version the alerts name. The published 0.9.103 image was built at v1.83.1, so this release is what clears both alerts and issue #967.

